PR #83, opened 2019-11-04
Project refactoring

tiulpin:
### A small suggestion:
- optimized imports
- updated requirements.txt
- removed duplicates from .gitignore
- refactored some *.sh scripts

Btw, would you like to use black formatter? (https://github.com/psf/black)
I can do a black PR by myself 😃


tiulpin:
Seems like no one is interested, OK)

PR #182, opened 2020-04-03
Validates dataset.json against schema file

ahmedhosny:
Example [here](https://github.com/MIC-DKFZ/nnUNet/blob/master/documentation/dataset_conversion.md) validates against schema.

Uses jsonschema which doesn't come out of the box with python, but I found it in my container (a dependency to some other lib required here). Can also be added it to requirements.txt for completeness..

### DIFF
--- a/documentation/dataset_conversion.md
+++ b/documentation/dataset_conversion.md
@@ -132,7 +132,7 @@ Here is the content of the dataset.json from the Prostate task:
      "description": "Prostate transitional zone and peripheral zone segmentation",
      "reference": "Radboud University, Nijmegen Medical Centre",
      "licence":"CC-BY-SA 4.0",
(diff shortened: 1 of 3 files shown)


FabianIsensee:
Hi there, thank you for this pull request! I will look into it this Friday :-)
Best,
Fabian

FabianIsensee:
Apologies for taking longer. Lots of things to do. I have not forgotten about it!

GregorKoehler:
Hi @ahmedhosny,
on behalf of all nnUNet maintainers, we want to apologize for leaving this issue stale for so long! In the meantime, nnUNet v2 came out. Are you working in v2 by now? And if so, would you say such a json schema file would still be beneficial here?

PR #189, opened 2020-04-08
Update setting_up_paths.md

arthursw:
dataset_conversion.md link is broken, probably because file structure has changed : 
`dataset_conversion/readme.md`  is now `dataset_conversion.md`


FabianIsensee:
Hi there, thank you so much for pointing out these deficiencies. nnunet is not hosted on Github but on our internal server. It is merely mirrored here. Merging pull requests from Github is not straightforward and rather complicated and I would like to do this only if there is a substantial change. Would it be OK for you if I just corrected these small errors on the master branch myself? For small changes it is best to open an issue and I will then make sure this is fixed.
Best,
Fabian

PR #190, opened 2020-04-08
Update dataset_conversion.md

arthursw:


FabianIsensee:
Hi there, thank you so much for pointing out these deficiencies. nnunet is not hosted on Github but on our internal server. It is merely mirrored here. Merging pull requests from Github is not straightforward and rather complicated and I would like to do this only if there is a substantial change. Would it be OK for you if I just corrected these small errors on the master branch myself? For small changes it is best to open an issue and I will then make sure this is fixed.
Best,
Fabian

arthursw:
Hi Fabian, sure I'll open an issue if I see other small changes. Feel free to update however you want, I just wanted to mentioned those typos.

PR #401, opened 2020-11-25
added saving and reloading of best epoch and loss

steflbert:
I had issues when restarting my training. The training would restart, but it did not remember the best model and the second model after restarting would be used. I think this is because the corresponding state variables were not saved, which I added.


FabianIsensee:
I do not use the best model, which is why I probably never noticed this. Thanks

PR #628, opened 2021-05-07
Update utils.py

tommydino93:
If there are some test images, numTest should change as well (I guess) :)


FabianIsensee:
Hi,
thanks for pointing my attention to it. I have resolved the problem in a different way (simply use `len(test_identifiers)`).
Best,
Fabian

PR #643, opened 2021-05-23
Change default number of multiprocessing cores

marcus-wirtz-snkeos:
Hi @FabianIsensee 
after having multiple crashes on our cluster (submission to different types of machines) due to the default value of 8 cores independently on the machine. I propose this way to use as default one core less than the number of physical cores (cf. here):
https://stackoverflow.com/questions/64837376/how-to-efficiently-run-multiple-pytorch-processes-models-at-once-traceback

It performed well across all machines for me.

Feel free to merge or discard it ;-)

### DIFF
--- a/nnunet/configuration.py
+++ b/nnunet/configuration.py
@@ -1,5 +1,8 @@
 import os
+import psutil
 
-default_num_threads = 8 if 'nnUNet_def_n_proc' not in os.environ else int(os.environ['nnUNet_def_n_proc'])
+default_num_threads = max(1, psutil.cpu_count(logical=False) - 1)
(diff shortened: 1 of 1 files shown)


FabianIsensee:
Hi @MRCWirtz I am happy to merge this but it would be great if default num threads never exceeded 8. If the number is too high there is a high probability that we run OOM for large datasets.

Do you want to adapt the number of threads used for data augmentation as well? That is covered somewhere else. I would however personally prefer that to remain as it is because this will give us problems on our cluster (environment variable `nnUNet_n_proc_DA` gives you control over that anyways)

Best,
Fabian

PR #703, opened 2021-07-21
Allow image modalities that are not numbered consecutively

tobihaui:
In my projects working with DIXON-MRI, I usually encounter 4 or 6 different modalities selecting different combinations for different tasks. Already trained models potentially need to adapt their plans.pkl file.

Changes:
Read dictionary items from modalities instead of range()-ing.
Allow using the same image dataset (e.g. DIXON MRI) for different tasks with
different focus and modality requirements.

### DIFF
--- a/nnunet/experiment_planning/experiment_planner_baseline_3DUNet.py
+++ b/nnunet/experiment_planning/experiment_planner_baseline_3DUNet.py
@@ -361,10 +361,10 @@ class ExperimentPlanner(object):
         modalities = self.dataset_properties['modalities']
         num_modalities = len(list(modalities.keys()))
 
-        for i in range(num_modalities):
-            if modalities[i] == "CT" or modalities[i] == 'ct':
+        for i, mod in modalities.items():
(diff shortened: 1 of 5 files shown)


FabianIsensee:
Hi, can you please specify how users are supposed to specify the modalities needed for a specific training? Is this done via manually editing the plans files?

tobihaui:
Hi, in case of Dixon-MRI, I used to specify the modalities _0000 = fat, _0001 = water, _0002 = in-phase, _0003=opp-phase. The goal is to keep the image files as they are when training a model using e.g. fat and opp-phase only.
In the current version of nnUNet, I had to rename the opp-phase image from _0003 to _0001 to work properly.

FabianIsensee:
In that case this looks more like a workaround for a very specific problem. I am not sure whether we should complicate things more for all the other users - the _0000 nomenclature is giving people headaches already

tobihaui:
Nothing would be changed in the _0000 nomenclature, but using the same data set to train different models is faciliated by keeping the same sequence number without renaming.

PR #852, opened 2021-11-15
Mofify early stopping for network_trainer and nnUNetTrainerV2

nntrongnghia:
nnUNetTrainerV2's behavior **before** modification:
- No early stopping
- Training stop at 1000 epochs
- Learning rate schedule follow an exponential function based on init_lr and actual epoch

nnUNetTrainerV2's behavior **after** modification:
- Max epoch is still 1000 epochs if no early stopping occurs
- If the validation criterion has not improved for 50 epochs, the learning rate is divided by 2.
- Learning rate schedule stays the same
- If the learning rate drops below 1e-6 and there is still no improvement for 50 epochs, the training ends.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -113,3 +113,4 @@ ENV/
 
 *.model
 
+nnunet/.vscode/launch.json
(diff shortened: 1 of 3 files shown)


FabianIsensee:
Hi,
I really appreciate your work on this and sorry for taking so long to react. We should leave nnUNetTrainerV2 and nnUNetTrainer as they are in order to stay compatible with the paper and its results. If you want to bring back early stopping then I would suggest you create a new trainer class with a different name where you implement these changes.
Best,
Fabian

PR #933, opened 2022-02-01
Correct list of pretrained models for BraTS

mzenk:
According to the paper, only region-based training was used, ensembling 5 + 5 + 15 models. I removed the checkpoints that seemed excessive.


mzenk:
Alright, makes sense.

Issue #607, opened 2021-04-20
"AssertionError: Labels must be in consecutive order" error when having more than 10 labels


paul-reiners:
I have more than 10 labels that are in consecutive order.  When I call `generate_dataset_json`, they are in consecutive order.  However, when they are saved to "dataset.json", they are converted to strings and saved in lexicographical order, like this:

        "0": "Unknown",
        "1": "Left-Cerebral-Exterior",
        "10": "Left-Thalamus-Proper*",
        "100": "Left-wm-intensity-abnormality",
        "1000": "ctx-lh-unknown",
        "10000": "unknown-10000",
        "10001": "unknown-10001",
        "10002": "unknown-10002",
        "10003": "unknown-10003",
        "10004": "unknown-10004",
        "10005": "unknown-10005",
        "10006": "unknown-10006",
        "10007": "unknown-10007",
        "10008": "unknown-10008",
        "10009": "unknown-10009",
        "1001": "ctx-lh-bankssts",
        .
        .
        .

Is there a fix for this or a workaround?

FabianIsensee:
wait - why do you have so many labels? This is unlikely to work at all.

Is it a problem if they are displayed in a different order?

Azumix:
Hey all,
Just discovering this issue working with nnU-Net. Why can't we just put the parameters `sort_keys` from the function `save_json` as a parameter of `generate_dataset_json`. That way, working with 10+ labels, you can just let the responsibility to the programmer to sort its dictionary before calling `generate_dataset_json`. Just tested it with my dataset, it does work and pass the sanity checks.

PR #1005, opened 2022-04-01
Add sort_keys parameter to generate_dataset_json

Azumix:
Hey everyone!
Here is my attempt to Fix MIC-DKFZ/nnUNet#607. With the addition of this parameter (which default value is set to `True` to stick with what is in place at the moment, you can choose to sort or not the keys of the label dictionary. Note that, if you set it to `False` it is your responsibility to unsure that your labels are in a consecutive order.

### DIFF
--- a/nnunet/dataset_conversion/utils.py
+++ b/nnunet/dataset_conversion/utils.py
@@ -25,7 +25,7 @@ def get_identifiers_from_splitted_files(folder: str):
 
 
 def generate_dataset_json(output_file: str, imagesTr_dir: str, imagesTs_dir: str, modalities: Tuple,
-                          labels: dict, dataset_name: str, license: str = "hands off!", dataset_description: str = "",
+                          labels: dict, dataset_name: str, sort_keys=True, license: str = "hands off!", dataset_description: str = "",
                           dataset_reference="", dataset_release='0.0'):
     """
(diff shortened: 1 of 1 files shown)


FabianIsensee:
How does this behave with more than 10 labels? If we support sorting then it should work all the time

Azumix:
Hello!
Well, the parameter just enable/disable the sorting feature. So with the new fix you deployed today, which allow keys in int order instead of string it will just sort the keys if the parameter if set to `True` and don't sort it if it is set to `False`. Before your fix it allowed me to sort my keys on my side then call your function.

FabianIsensee:
The keys will still be string. What I want to avoid is that if someone sorts their keys properly (1, 2, 3, 4, ..., 10, 11, 12) that this sorting will be messed up if sort_keys=True (1, 11, 12, 2, 3, 4, ....). This should not be the case

Azumix:
That's exactly what I also wanted to avoid, as my keys where correctly sorted directly, in the end it was messed up by the integrated sorting. 
Thus, it is what is allowed here, if you have your keys that are already well sorted you just pass `sort_keys=False`, then the function will trust you and won't try to sort the keys and thus, it won't mess up everything.

FabianIsensee:
ah now I see :-D

PR #1011, opened 2022-04-04
fix misjudgement of labels consecutity when the number of labels over 10

Yuanke-Pan:
When the number of labels is over 10,  it comes up with
AssertionError: Labels must be in consecutive order (0, 1, 2, ...). The labels [10  2] do not satisfy this restriction
sorted function can fix the problem

### DIFF
--- a/nnunet/preprocessing/sanity_checks.py
+++ b/nnunet/preprocessing/sanity_checks.py
@@ -166,7 +166,7 @@ def verify_dataset_integrity(folder):
 
     # verify that only properly declared values are present in the labels
     print("Verifying label values")
-    expected_labels = list(int(i) for i in dataset['labels'].keys())
+    expected_labels = sorted(list(int(i) for i in dataset['labels'].keys()))
 
     # check if labels are in consecutive order
     assert expected_labels[0] == 0, 'The first label must be 0 and maps to the background'


Azumix:
Hello! 
We had the same idea, I did propose a fix there: #1005 
Beside that a simple sort would work it is already implemented in the code. However, the parameter handling the sort is not used.

Yuanke-Pan:
Sorry, I didn't check the previous pull request and issue. I'm glad that we have the same idea. I have no means to offend, but I don't think adding the sort_keys parameter can solve #607 , because in the save_json function the sort_keys parameter is Ture by default, you can check it in another repository. I think this parameter just makes sure the keys are sorted by str order, rather than int order.

Azumix:
You're right saying that the `sort_keys` parameter only ensure that the keys are sorted but there as as str. Which is the problem and why we end up with "1,10,11" order. 

The idea behind my PR is, if you have less than 10 you can let everything by default, if you have more than ten, you unsure yourself that your keys are sorted and set `sort_keys` to `False`. That way, you will not end with "1,10,11".  But sure, it's then the user responsability to correctly sort its keys. 

And don't worry for offending, I'm always happy to discuss. :)

FabianIsensee:
Hi Yuan,
it seems like you have the labels 10 and 2. The labels need to be consecutive, starting with 0, so
0= background
1=class1
2=class2
Otherwise nnU-Net will not work, regardless of whether they are sorted or not. It actually doesn't matter whether the labels are sorted correctly in the dataset.json
Best,
Fabian

Yuanke-Pan:
Hi Fabian,
Actually, I have 13 labels in the dataset.json from 0 to 12, but I still failed in the consecutive check.  From the original code, it just gets the key from the dataset.json, turned the key to an integer, and applies `np.ediff1d` to the key list to make a judgment. When `np.ediff1d` equals one it passes, otherwise it fails.  If the consecutive labels are not sorted, 'np.ediff1d' will not equal one.

Yuanke-Pan:
That's OK, please never mind it. I will try to make a pull request more noticeable next time. :-P
Best,
Yuan

FabianIsensee:
That's not on you but on me. Whenever I find the time I just go to the github folder in my mails and open everything that came in since the last time. Then I have issues mixed with PRs and that was what got me mixedup

PR #1013, opened 2022-04-05
Remove mentions of --strict flag from readme

bartosz-grabowski:
`--strict` flag is no longer used in `nnUNet_find_best_configuration`


FabianIsensee:
thanks!

PR #1113, opened 2022-07-22
introduce max_num_epochs as command line argument

chrisrapson:
`max_num_epochs` is something that I'd like to be able to customise. Various github issues, and the existence of classes specifically for 2 or 5 epochs seem to confirm that, so I've added it as a command line option.

The functional changes are in `run_training.py` and `nnUNetTrainerV2.py`. Everywhere else is plumbing to make the signatures match.

The default is still 1000 most places. Where `max_num_epochs` was set to something else, I've used the a default which matched that class' existing `max_num_epochs`, but then allowed it to be ignored/overwritten. Other options that might be more logical could be:

- an assert, to make sure that if you're calling the `nnUNetTrainerV2_5epochs`, then `max_num_epochs` should be `5`
- removing those classes, since they are redundant now

In order to keep the same signature for `nnUNetTrainer` and `nnUNetTrainerV2`, I've added `max_num_epochs` as a parameter to both, even though it has no functionality for `nnUNetTrainer`.

Ideally, this change should also be reflected in an update to the documentation, but I don't think that's urgent, since there should be no change in functionality for users who are not aware of it.


osbm:
This is definitely something we need. Also, we need every other training option to be an argument. Instead of thousands of classes.

FabianIsensee:
Hey, this contradicts the idea behind nnU-Nets data structure where the name of the trainer is used in the name of the folders. We have nothing in place to keep track of hyperparameters that are changed at run time and it would create a mess for everyone who is working with a larger number of experiments. I would therefore like to keep things as they are. I hope you understand

PR #1155, opened 2022-09-01
ENH: Add multithreading support on the windows platform

rbumm:
as suggested by @CedarYang


rbumm:
With this fix, I am able to run nnUnet on Windows 11 in a Powershell.

rbumm:
Could I bring this PR up again? It is a simple fix that enables nnU-Net in Windows, but finding and changing the file can be challenging for newbies.

lassoan:
Please review&merge. This is a very low-risk change and allows prediction on Windows both using CPU and GPU.

PR #1207, opened 2022-10-27
In DDP mode, Fix the error of deleting the latest model at the end of training 

lvguofeng:
In DDP mode, at the end of training, different processes will delete the latest model, causing some processes to enter the if statement, but when "os.remove" is executed, other processes have deleted the file, so an error "FileNotFoundError: [Errno 2 ] No such file or directory" will be reported.

Refer to the methods "save_checkpoint", "plot_progress", write the delete model as a function, under DDP, rewrite this function, just add "if self.local_rank == 0"


FabianIsensee:
This issue has already been resolved in a different way, nonetheless thanks a lot for your PR!

PR #1209, opened 2022-10-28
Require `scikit-learn` instead of `sklearn`

StellarStorm:
Hi @FabianIsensee - thanks for the wonderful project!

I noticed that `sklearn` is required and installed by pip even when `scikit-learn` is already installed on the system. Looking on PyPI, `sklearn` seems to be a [dummy package](https://pypi.org/project/sklearn/) that says to use [`scikit-learn`](https://pypi.org/project/scikit-learn/) instead. This just updates the requirements so we install the correct package directly.

For reference, `sklearn` was added to the requirements file in commit 6b5419a57023c0ca8b909c491c55d569ed886baf

Thanks!


denbonte:
From December 2022, in theory `pip install sklearn` should straight up fail (for me, it doesn't happen 100% of the times - but it might be down to other factors, e.g., what else I have in the containers I'm installing `nnunet` in).

Still - I think this pull request is a must (as a lot of new users could have problems installing the package otherwise)!

hroskes:
Would it be possible to merge this, please? It's a trivial fix for a problem that is causing me issues in building my docker containers.

denbonte:
> Would it be possible to merge this, please? It's a trivial fix for a problem that is causing me issues in building my docker containers.

@hroskes in the meantime, you can try with:

```
ENV SKLEARN_ALLOW_DEPRECATED_SKLEARN_PACKAGE_INSTALL=True
```

before pulling `nnunet` or any of the packages based on it. Ugly, but works for me.

piiq:
Hey @FabianIsensee 👋
Is there anything missing or anything I can do for this to be merged?

lwjason:
The issue becomes more often in now. Every first 10 mins of each hour are failing. It would be great to see this tiny change merged :)

FabianIsensee:
Sorry people, busy times :-) I am cathcing up now

hroskes:
Thank you!

PR #1369, opened 2023-04-01
Fix attempting to load non-existent checkpoint

ramtingh:
Based on "Starting a new training..." I assume the intent is for this not to be a fatal error (which is different from nnUNet v1). But since `expected_checkpoint_file` isn't reset it still crashes attempting to read the non-existent checkpoint_best.pth file.


FabianIsensee:
thanks, makes sense!

PR #1372, opened 2023-04-01
Add refernce to nnUNetv2 to readme.md

Nils-ChristianIseke:
Added a reference to nnUNetv2.
This is because I implemented everything for nnUnet and only afterwards realized that nnUNetv2 exists.


FabianIsensee:
thanks, good idea!

PR #1404, opened 2023-04-18
Fix incorrect type bug

Rpeng666:
The folds variable is parsed from command `-f xxx` . The possible values are numbers (0,1,2,3,4) or strings ('all'), but they are all parsed as strings. So when it's a number, `["fold_%d" % i if i ! = 'all' else 'fold_all' for i in folds]` will report an error. Just change it to `"%s"`


FabianIsensee:
thanks!

PR #1414, opened 2023-04-21
Default num processes fix

joshestein:
A few small fixes:

1. The documentation for the default num processes for "2d" configuration species the num processes as 8, but was defaulting to 4.
2. Make the order of default num processes match the default order of the '-c' option (2d, 3d_fullres, 3d_lowres).
3. Since a set was used to store `np`, it would not allow duplicate num procceses. If we had (8, 4, 8), this would reduce to (8, 4), and there would subsequently be an [error thrown due to mismatching number of processes and number of configurations.](https://github.com/MIC-DKFZ/nnUNet/blob/master/nnunetv2/experiment_planning/plan_and_preprocess_api.py#L95)


FabianIsensee:
thanks!

PR #1415, opened 2023-04-21
ACDC dataset conversion helper

joshestein:
Adds a conversion helper for the Automated Cardiac Diagnosis Challenge (ACDC) dataset.

The dataset ID was chosen as 27 for consistency with the nnUNetv1 helper.


FabianIsensee:
Thanks!

PR #1440, opened 2023-05-09
Using numpy allclose instead of isclose().all()

ancestor-mithril:
There is a (small) performance benefit when using `np.allclose` even when the arrays are equal.


FabianIsensee:
thanks!

PR #1443, opened 2023-05-10
Move package metadata to `pyproject.toml`

MathijsdeBoer:
Modern Python development is starting to move to a unified [`pyproject.toml`](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/) file as opposed to the various ways of defining package metadata.

This pull request replaces the old style `setup.cfg` and `setup.py` to this `pyproject.toml` format.
Functionality has not changed. `pip install -e .` still works as expected, as do the entrypoints:

```bash
(venv) PS D:\PythonProjects\nnUNet> nnUNetv2_train    
nnUNet_raw is not defined and nnU-Net can only be used on data for which preprocessed (...)
(...)
```

I've also added an example set of optional developer dependencies, which can be installed using the square brackets at the end of the package name.

For local install:

```bash
(venv) PS D:\PythonProjects\nnUNet> pip install -e .[dev]
```


GregorKoehler:
Hi @MathijsdeBoer, 
on behalf of all nnUNet maintainers, we want to apologize for leaving this issue stale for so long! This change seems interesting. Googling this, I read that for editable installs the setup.py file is still required? Is this still the case? Also, I wonder where the benefits are when using the .toml format? Maybe you can share your insights here?

MathijsdeBoer:
Hey @GregorKoehler!

No worries about the delay, I must admit I kind of forgot about this as well!

The `pyproject.toml` file is [meant as a replacement of the `setup.py` files](https://peps.python.org/pep-0518/#rationale), mainly as a common standard between build systems. This means that you don't need the `setup.py` file to run `pip install -e .` for editable installs. Something I've done across a few different projects myself.

While I expect the `setup.py` file to be accepted for a good while to come, the `pyproject.toml` file is a bit more flexible in what you can store in it; development packages like Black, Isort and Ruff (and probably more) are configurable in this file, removing the need for many separate config files. On top if this, on a more subjective front, script entrypoints, (optional) dependencies, metadata and deployment gets simplified, as it removes a few assumptions that common Python tooling has historically had to make.

I hope I've covered most of the questions you had! I fully admit this change would mostly be a nice-to-have (or at least untill the unlikely sunsetting of the `setup.py` system), rather than a must-have, so I won't be too disappointed if this doesn't go through.

FabianIsensee:
Thanks a lot! I had to make a couple of changes because the PR wasn't installable in an editable way (which I constantly use)

MathijsdeBoer:
Interesting that it didn't work out of the box for you, as it worked without the `setup.py` for me! As have several other projects that I've used the same setup for.

Either way, glad to help!

PR #1447, opened 2023-05-11
Using torch.sum over multiple dimensions

ancestor-mithril:
Summing a tensor over multiple dimensions was added in https://github.com/pytorch/pytorch/pull/6152. 

The error between the two methods is arround `1e-6` for me, but I guess this is ok and expected.  
```py
error = 0.0
for i in range(100):
    net_output = torch.rand((10, 5, 3, 10, 10)).cuda()
    axes = tuple(range(2, len(net_output.size())))

    r1 = sum_tensor(net_output, axes, keepdim=False)
    r2 = net_output.sum(dim=axes, keepdim=False)
    error += (r1 - r2).abs().mean().item()
print(error / 100)
```


FabianIsensee:
Yeah that sum_tensor was a relic form when torch didn't support this. Thanks!

FabianIsensee:
Your pull request is merged and your contribution is visible in the git history. Github just doesn't see that in this thread because the merge happened in our internal gitlab instance. Thanks again!

PR #1476, opened 2023-05-31
Pathology

nadieh:
weight and biases is added


FabianIsensee:
Hey you seem to want to merge a lot of your custom pathology related implementations. I think it would be better to do this in a separate repository (which is a fork of nnU-Net) because we need to be able to maintain the things that are in here. I do not feel comfortable that we can do that. 
Due to the issue of maintainability I would also prefer not to integrate wandb into the main repository. What we could do though is create a special branch where your wandb integration is added. For that you would need to isolate the wandb part of your PR through. Would you like to do that?

PR #1503, opened 2023-06-19
There are some ambiguous descriptions in region_based_training.md

SherlockZYB:
The numbers in the 'region_class_order' are not the order in which the labels are written, but which number of labels the regions correspond to!
It's easy to misunderstand when seeing the name 'region_class_order' for the first time, I just thought that it means the order how will the labels be written, on the other hand, I thought the order of the numbers in the list will determine the representations of output labels, which means: '1' is the **first** number in the list, so it corresponds to the **label 1**. The given example is still right even if I understand it like this, and this can result in outputs different from the ground truth. So it's essential to change the example a little. 
I have used 3 days to find out why label 1 and label 2 are exchanged in my experiments because the key code(label_handling.py, line 171) is in starmap_async(), and the breakpoint debugging will jump it. T T

### DIFF
--- a/documentation/region_based_training.md
+++ b/documentation/region_based_training.md
@@ -54,13 +54,14 @@ For region-based training, the labels need to be changed to the following:
     "tumor_core": [2, 3],
     "enhancing_tumor": 3  # or [3]
 },
-"regions_class_order": [1, 2, 3],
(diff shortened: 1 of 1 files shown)


FabianIsensee:
Thanks a lot! Changes are merged (and corrected a bit ;-) )

PR #1586, opened 2023-07-29
Update default_preprocessor.py

cuchoco:
I was confused because it says 3D in the annotation, when in a 2D configuration it doesn't convert the spacing between slices like it does in 3D.


FabianIsensee:
merged, thanks!

PR #1591, opened 2023-08-02
Update map_channel_name_to_normalization.py

dhaberl:
Hello Fabian,

There is an inconsistency between the documentation (explanation_normalization.md) and the code in the mapping of the different normalization schemes.

Without the proposed change, 'rescale_to_0_1' in the 'channel_names' of 'dataset.json' will trigger 'ZScoreNormalization' instead of 'RescaleTo01Normalization'.

Best,
David


FabianIsensee:
good catch, thanks!

PR #1604, opened 2023-08-08
Delete LibreOffice's lockfile

moschams:
LibreOffice is awesome! But we may not need lockfile in the root directory.


moschams:
Due to the lockfile being deleted, this PR was closed.

PR #1630, opened 2023-08-17
Zero Grad with set_to_none True

ancestor-mithril:
Reference: https://pytorch.org/docs/stable/generated/torch.optim.Optimizer.zero_grad.html. 
Set to none can be safely used in nnUNet.


FabianIsensee:
thanks a lot! Doesn't do much for run time but saves a little bit of GPU memory
Task03_Liver from 7800 to 7500 MB

PR #1662, opened 2023-08-30
Convert to f-strings

joshestein:
Use [flynt](https://github.com/ikamensh/flynt) to convert formatted strings to f-strings. In general, these are more readable.


FabianIsensee:
thanks a lot! Those were relics from the old nnU-Net days :-) Glad this is cleaned up now

PR #1676, opened 2023-09-04
Add missing character in `pyproject.toml`

MathijsdeBoer:
This probably happened when I was moving the entrypoints to the `pyproject.toml` file, but `nnUNetv2_export_model_to_zip` was missing a `y` at the end of the line, causing this command to not run.


FabianIsensee:
Hey, this should already be fixed. Sorry I didn't see your PR before :-/

PR #1711, opened 2023-10-03
Using tuple equality instead of checking element by element

ancestor-mithril:
Python supports tuple and list equality and it is more clear and slightly faster (if this matters) than `all([a == b for a, b in zip(x, y)]`.
```py
>>> (1, 2, 3) == (1, 2, 3)
True
>>> (1, 2, 3) == (1, 2,)
False
>>> [1, 2, 3] == [1, 2, 3]
True
>>> [1, 2, 3] == [1, 2]
False
>>> [1, 2] == (1, 2)
False
```


FabianIsensee:
thanks a lot, I seem to have become lazy :-)

PR #1720, opened 2023-10-04
Turning on compression for SimpleITK writer.

yotamyaniv:
When saving segmentation in nrrd format without compression the file size is significantly larger than with compression, 20mb versus 19kb for a 5000x4000 nrrd binary segmentation.


FabianIsensee:
thanks a lot. Never noticed this because for me it's nifti all the way :-)

PR #1721, opened 2023-10-05
add codespell workflow (to prevent future typos), config and get some typos fixed

yarikoptic:
since `split` is not any kind of a special keyword I decided to agree with codespell to just use it.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,22 @@
+---
+name: Codespell
+
+on:
+  push:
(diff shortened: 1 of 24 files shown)


FabianIsensee:
Hey, thanks a lot. I am not familiar with codespell - how will it let me know about future mishaps?

yarikoptic:
> Hey, thanks a lot. I am not familiar with codespell - how will it let me know about future mishaps?

Workflow should error out if new typo is found.

PR #1734, opened 2023-10-10
Using `@cached_property` to cache `@property` calls.

ancestor-mithril:
@cached_property works like `@property` combined with @lru_cache, but it caches the result only once and it assumes it never changes (the internal state of the class is not modified in the meantime). 
I also assume that the internal state of `LabelManager` and `ConfigurationManager` does not change after creation, please correct me if I'm wrong.


FabianIsensee:
Hey I would prefer to leave it as is because should I ever decide to fiddle with these classes having @cached_property might cause some really obscure bugs that will be difficult to find

PR #1740, opened 2023-10-11
Reworked deep supervision

TaWald:
Currently deepsupervision is half-static in the code and is annoying to turn on/off. 
I introduced `self.enable_deep_supervision` which is now a attribute, accessible in the class methods.

The deep_supervision flag itself is used at different positions:
1.  `def _build_loss`, which is a class method, that (currently always) wraps the loss with a deep supervision wrapper. Now instead of wrapping always it checks if the flag is set and only wraps then.
2.  `def build_network_architecture`. Here it's hard baked and one cannot change it. If one would like to make a non_DS architecture one would have to overload the entire initialize 💩 -- now its using the class attribute
3. `def _get_deep_supervision_scales ` This used to always output scales of the DS, now it's conditionally outputting either the scales, or `None` (which is relevant for (4)
4. `def get_training_transforms` (and validation transforms) here one has to pass None for the DataAug to not return a list of labels on the wanted scale, originally one would have to overload the` _get_deep_supervision_scales` which is unnecessary now.
5. `def on_train_start(self):` uses a fixed value to set `ddp` to deep supervision, now depends on attribute
6. `def validation_step` used to always take the highest resolution (of the list), now it's conditional so no overloading of the validation step when one uses DS or not

Now the `NoDeepSupervisionTrainer` is basically only setting deep_supervision to False, which imo is way cleaner than before, but you can judge for yourself.  Also all the downstream Trainers that changed the above mentioned functions were updated to be compatible to the new structure. (also since basically all changes are (`if self.enable_deep_supervision: [...]` it should marginally influence systembehavior when left at default value of true in`__init__` 

(P.S. When changing the files black formatted automatically, hence it looks like more changes than it actually is.)


FabianIsensee:
Thanks, nice work!

PR #1756, opened 2023-10-18
Refactor _internal_maybe_mirror_and_predict function

Plasma-Blue:
This PR replaced the multiple if statements with a single loop that uses itertools.combinations to generate all possible axes combinations for mirroring. This reduces code duplication and (hopefully😉) improves readability and maintainability.


FabianIsensee:
Thanks! this is indeed cleaner

PR #1826, opened 2023-12-01
Add ONNX exporter

MathijsdeBoer:
# Draft

This pull request is still a draft, pending some quick testing. Therefore, it currently serves as a gauge of interest from the current maintainers.

# Purpose

Hey there! I, and others, have occasionally wondered if it was possible to export the trained nnU-Nets to an ONNX format, for deployment, or other sharing. To this end, I've cobbled together an exporter using the built-in pytorch ONNX exporter.

In the past, I had managed to get [an exporter working for v1](#1144), but this won't work since the release of v2 changed some things around. Nevertheless, the exported information let us build a very minimalist copy of the entire nnU-Net pipeline that produced outputs faster than the v1 inference pipeline, and with less overhead.

# Implementation

I've "repurposed" bits and pieces of code I found around the project (Mainly `nnunetv2.inference.predict_from_raw_data.py`) to load in a network, set the parameters and acquire any information the user might want to build their own ONNX pipeline. The command works roughly the same as the `nnUNetv2_export_model_to_zip` command, but exports an `.onnx` file for each fold and configuration combination instead. Each `.onnx` file should be accompanied with a `.json` file that has some basic information the user might want to know to build their own ONNX pipeline with.

Finally, it's made perfectly clear that any ONNX pipelining is the sole responsibility of the end-user, not the maintainers of nnU-Net through a big warning box which is printed before the export begins.

# Motivation

While I can't speak for any others' reasons, our main reason to want an ONNX format is the deployment of the trained models to an inference server. Due to server costs, we might want to use some of ONNX's excellent optimization techniques for faster inference, maybe move away from a python-based pipeline in favor of compiled languages, like C. C++ or Rust. Finally, a lot of CUDA+Python Docker containers produce gigantic files (for example, the docker image I use to train the nnU-Nets is 24GB when exported), which will take an extremely long time to boot up on servers.

A lot of clinicians really would like to see neural networks applied in their practice, and quite a few companies are willing make that investment. However, server costs are still a large burden, so any ability to offer a way to export an nnU-Net to a common format, even without further support beyond the exporter, will hopefully allow a larger adoption outside of pure research across the board.

### DIFF
--- a/nnunetv2/model_sharing/entry_points.py
+++ b/nnunetv2/model_sharing/entry_points.py
@@ -1,6 +1,9 @@
+from pathlib import Path
+
 from nnunetv2.model_sharing.model_download import download_and_install_from_url
 from nnunetv2.model_sharing.model_export import export_pretrained_model
 from nnunetv2.model_sharing.model_import import install_model_from_zip_file
+from nnunetv2.model_sharing.onnx_export import export_onnx_model
 
 
 def print_license_warning():
@@ -51,7 +54,7 @@ def export_pretrained_model_entry():
     parser.add_argument('-p', required=False, type=str, default='nnUNetPlans', help='plans identifier')
     parser.add_argument('-f', required=False, nargs='+', type=str, default=(0, 1, 2, 3, 4), help='list of fold ids')
     parser.add_argument('-chk', required=False, nargs='+', type=str, default=('checkpoint_final.pth', ),
-                        help='Lis tof checkpoint names to export. Default: checkpoint_final.pth')
+                        help='List of checkpoint names to export. Default: checkpoint_final.pth')
     parser.add_argument('--not_strict', action='store_false', default=False, required=False, help='Set this to allow missing folds and/or configurations')
     parser.add_argument('--exp_cv_preds', action='store_true', required=False, help='Set this to export the cross-validation predictions as well')
     args = parser.parse_args()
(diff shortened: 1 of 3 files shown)


thangngoc89:
@MathijsdeBoer thank you for a very detailed answer.

> Python, PyTorch and nnUNet are only requirements for the export step, but not beyond that. It isn't my intention to move nnUNet away from PyTorch with this PR, as it offers a lot of benefits for training and ease of use. My only intention is for users to be able to easily export their models to do with as they see fit, beyond the scope of the nnUNet project.

Definitely. I was talking about not having Pytorch on inference mode.

I will try to test this PR locally and provide feedbacks soon

FabianIsensee:
Hey @MathijsdeBoer thank you so much for all this work! Exporting to onnx is certainly something we should support. One of the reasons I have never done that so far is because I don't want to be responsible for people not being able to reproduce our results with their onnx files. After all, they would need to reimplement a lot of things around the model: Cropping, resampling, intensity normalization, sliding window inference with proper logit aggregation, resizing of the probabilities to the original data shape, conversion to segmentation, revert cropping. There are many opportunities for errors.

I am happy to accept this PR once you are happy with the config.json file etc. Just ping me when everything is complete.

One more thing (and please let me know if that doesn't make any sense - I am not very familiar with onnx and optimizations): Shouldn't it be possible to use the onnx optimized model in the regular nnU-Net inference code as well? Have you experimented with that?
Best,
Fabian

MathijsdeBoer:
Hey Fabian, no worries. It didn't end up being a massive amount of work, as PyTorch has a built-in exporter! Main time was spent reading through documentation and nnUNet core code to see if I could reuse as many existing functions as possible. That should make it so that any updates to the core code will keep the exporter up to date as well. Barring any fundamental rewrites, of course.

> One of the reasons I have never done that so far is because I don't want to be responsible for people not being able to reproduce our results with their onnx files.

Yep, I recall seeing you mention something like that a few times before. I could make the warning text a lot stronger, and add a disclaimer for any future support. Maybe something like:

```text
######################################################
!!!!!!!!!!!!!!!!!!!!!!!!WARNING!!!!!!!!!!!!!!!!!!!!!!!
######################################################
Exported models are provided as-is, without any
guarantees, warranties and/or support from MIC-DKFZ,
any associated persons and/or other entities.

You will bear sole responsibility for the proper
use of the exported models.

You are responsible for creating and validating
the ONNX pipeline yourself. To this end we provide
the .onnx file, and a config.json containing any
details you might need.
######################################################
```

Of course, there's _always_ going to be someone who might ask questions on this, but this will at least make it pretty clear that they shouldn't expect any hand holding.

One thing that might be nice, however, is a short document that has a step-wise overview of how the inference works. Optionally with where to find these things in the code. Just as a general reference for pipeline implementations.

As for being happy with the config file, I think I've managed to include everything someone might need to accurately rebuild the pipeline with all possible export configurations. The only thing I haven't been able to test is what happens if someone has a label with multiple values. Fortunately, given the free nature of onnx pipeline implementations, there will be a lot of room for hardcoding things to fit their particular needs, so it might not be a big problem after all.

I think that if you want to use ONNX in the existing nnUNet pipeline, you'd have to decouple the `nnUNetPredictor` from the `nn.Module`, and simplify it down to basic steps. ONNX doesn't rely on `torch.Tensor` for data, and uses `numpy.ndarray` as input and output instead. Other than that, general steps are very similar!

```python
...
# PyTorch
data = preprocess_data()
data = torch.tensor(data)
data.to(device)

torch_model = get_model()
torch_model.eval()
with torch.no_grad():
    pred = torch_model(data)

pred.detach().cpu().numpy()
postprocess_pred(pred)

...
# ONNX
data = preprocess_data()

ort_model = ort.InferenceSession(
    model_filepath,
    providers=["CUDAExecutionProvider"]  # Falls back to CPU automatically
)

# ONNX models use named inputs/outputs, exporter simply calls it "input", but this is more general
ort_inputs = {ort_model.get_inputs()[0].name: data}
# Identical to following in our case
ort_inputs = {"input": data}

pred = ort_model.run(None, ort_inputs)[0]

postprocess_pred(pred)

...
```

There's a few more things one can do before starting inference sessions, [such as offline optimizations ahead of time](https://onnxruntime.ai/docs/performance/model-optimizations/). This should make the model faster and/or smaller, but unfortunately might still impart a slight performance penalty. So these tradeoffs would be up to the end-user.

Sharpz7:
Hey @MathijsdeBoer,

I was wondering where you had gotten with this? Happy to give a helping hand with merge conflicts if you are short of time.

As well as that, a technical question - do you have the ability to merge the folds into a single ensemble model before exporting to onnx?

thangngoc89:
@Sharpz7 

> Currently I use normalisation as input_image = (input_image - input_image.mean()) / max(input_image.std(), 1e-8)

You can see normalization explanation here https://github.com/MIC-DKFZ/nnUNet/blob/master/documentation/explanation_normalization.md

The 0.5 and 99.5 percentile cutoff values could be extracted from the plan files

Sharpz7:
Thanks @thangngoc89 

@FabianIsensee, at least for me, the reason I am going along this route is as a ML Engineer trying to implement this work, I have my own pre-processing and post-processing of the data, so having 2x of each, with some fully integrated into your pipeline, and mine not, just adds to the complexity. If you can think of a better solution for me, then let me know. 

I have been unable to reproduce native results, so I am going to ditch my attempts for now (I got as far as being able to prove I have the pre-processing replicated up to the normalisation stage) , but I imagine I am going to run into more headaches getting nnunet integrated into a wider system anyways, so I might be back :))

Would be great if we can get this PR up with an example, at least for the default preprocessing and postprocessing workflows (i.e znorm + crop + pad for when window is larger than cropped image)

chaitanya-kolluru-elucid:
I have a related issue. I'm trying to [export the model](https://github.com/MIC-DKFZ/nnUNet/issues/2088) as a single torchscript .pt file first and then convert it to ONNX. I set the network to eval() before exporting to the torchscript model. However, I get this user warning when I run `torch.onnx.export`

```/usr/local/lib/python3.10/dist-packages/torch/onnx/symbolic_helper.py:1531: UserWarning: ONNX export mode is set to TrainingMode.EVAL, but operator 'instance_norm' is set to train=True. Exporting with train=True.```

I see some discussion about the instance_norm layer above, wondering if someone's had some luck with this issue. I looked into [some discussions on the pytorch repo](https://github.com/pytorch/pytorch/issues/94604), but didn't get too far based on their suggestions.

PR #1830, opened 2023-12-05
Minor update to documentation with information on TIFF spacing json file, resolves #1829

### LINKED ISSUE #1829: Documentation unclear about spacing json file that needs to be supplied
In the documentation:

https://github.com/MIC-DKFZ/nnUNet/blob/75c0e4608f8e9fcc52dba4b8c70f879a5b57aaa5/documentation/dataset_format.md?plain=1#L64

It says:

>nnU-Net expects each TIF file to be accompanied by an identically named .json file that contains three numbers 
(no units, no comma. Just separated by whitespace), one for each dimension.

Having tried a few combinations of things, none of which worked. I hit upon the following format for the json file.

```
{
    "spacing": [7.6, 7.6, 80.0]
}
```

And then set up the following structure inside one dataset folder:
```
./labelsTr/cell6.json
./labelsTr/cell6.tif
./imagesTr/cell6_0000.tif
./imagesTr/cell6.json
./dataset.json
```

This seemed to work. At least the spacing changed from [1 1 1]. I think the documentation could be clearer on this point. I didn't want to do a PR for this but I'm happy to, if you agree.

quantixed:


FabianIsensee:
thanks a lot!

Issue #1832, opened 2023-12-05
pip install -e . fails to install: results in UNKNOWN-0.0.0


mike-barrow:
pulling the repository, navigating to nnUNet  and running ```pip install -e .``` does not install nnUNet v2 in my virtual environment. Instead I see:
```
pip install -e .
Obtaining file:///home/king/workspace/Sergios-PET-Lesion-Seg-venvii/workspace/nnUNet
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Installing collected packages: UNKNOWN
  Running setup.py develop for UNKNOWN
Successfully installed UNKNOWN-0.0.0
```

Issue occurs on a virtual environment on Ubuntu 20.04 LTS, python version 3.10.12

Nils-ChristianIseke:
I had the same issue once. It was resolved by running `'pip install --upgrade pip setuptools`, before installing nnunet

mrokuss:
Hey @mike-barrow 

Could updating `setuptools` resolve your issue as in the case of @Nils-ChristianIseke ?

mike-barrow:
Hi sorry for the late response. No, `pip install -U setuptools` does not seem to work as it did before

mike-barrow:
Hi there, my mistake. I did not update setuptools correctly. I needed to specify pip3:
```
pip3 install --upgrade setuptools pip
python -m pip install .
```
after doing this, the install worked.

MikasaLee:
Hi, i also have this problem,`pip install -U setuptools`  and `python -m pip install .` does not seem to work

PR #1850, opened 2023-12-13
Adding build-system to pyproject.toml

Nils-ChristianIseke:
If using an older version of setuptools, the installation of nnunet fails (#1832). Adding the requirement of using setuptoosl >=69.0.2 should resolve this issue.

I had no time to test it yet, and it might also work with lower version of setup_tools than 69.0.2. Before merging this should be tested.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,6 +86,10 @@ dev = [
     "pre-commit"
 ]
 
+[build-system]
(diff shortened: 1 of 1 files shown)


FabianIsensee:
Thanks! 
67.8.0 also works so this is what I set as the minimum requirement

PR #1889, opened 2024-01-11
Improving normalization by reducing memory allocations

ancestor-mithril:
* using `np.ndarray.astype(..., copy=False)` to prevent new array creation when the types are the same
* using inplace numpy matrix operations
* using inplace `np.clip`

This version is up to 20x times faster for big `(512, 512, 512)` CT cases.


FabianIsensee:
thanks a lot!

PR #1943, opened 2024-02-10
Enable user to specify torch device in run_training

bdzyubak:
And stop overriding it with index 0 in the actual trainer


FabianIsensee:
Hey, what would you say is the advantage of doing it this way rather than using the environment variable?

Issue #1951, opened 2024-02-13
Clarify that nnUNet's internal data format uses `SimpleITK` axis order, requiring an axis swap for `nibabel`-loaded images


joshuacwnewton:
Hi there! I come from a lab that primarily uses `nibabel` over `SimpleITK` for image processing.

When using nnUNetV2, we did the following:

- Training: `nnunetv2_train` on `.nii.gz` images (no setting of `overwrite_image_reader_writer` in the `dataset.json` file) 
   - `SimpleITKIO` was implicitly used as the image reader/writer by default
- Inference: Load images in our Python script using `nibabel`, then pass data array to `predict_single_npy_array`

When following this approach, we learned of a quirk in [SimpleITK's behavior](http://insightsoftwareconsortium.github.io/SimpleITK-Notebooks/Python_html/03_Image_Details.html#Conversion-between-numpy-and-SimpleITK):

> SimpleITK and numpy indexing access is in opposite order!
> 
> SimpleITK: image[x,y,z]
> numpy: image_numpy_array[z,y,x]

Because of this, the x and z axes get swapped within `nnUNet` at this call: 

https://github.com/MIC-DKFZ/nnUNet/blob/997804c7510634dc8fd83f1194b434c60815a93e/nnunetv2/imageio/simpleitk_reader_writer.py#L41

As a result, when loading a data array using `nibabel` (for inference purposes), the array (and the spacings!) must be transposed (`[2, 1, 0]`) to match the format of the training data. In fact, we had no idea that the training data had been transposed by SimpleITK in the first place! (In essence, even though our training data was in LPI orientation, the model was effectively trained using IPL orientation due to the axis swap!)

It might be good to emphasize in the relevant documentation ([1](https://github.com/MIC-DKFZ/nnUNet/blob/master/documentation/dataset_format.md#supported-file-formats), [2](https://github.com/MIC-DKFZ/nnUNet/blob/master/documentation/dataset_format.md#datasetjson)) ~the priority difference for `.nii.gz` image readers (SimpleITK > nibabel), as well as the data loading differences between the `SimpleITKIO` loader and the `NibabelIO` loader~. (EDIT: I now realize that `NibabelIO` will result in the [same axis-swapping](https://github.com/MIC-DKFZ/nnUNet/blob/997804c7510634dc8fd83f1194b434c60815a93e/nnunetv2/imageio/nibabel_reader_writer.py#L54-L55), making the choice of reader not as important as I thought.) 

[...]

PR #2033, opened 2024-03-23
Clarify usage of `predict_single_npy_array` re: SimpleITK axis ordering

joshuacwnewton:
## Description

This PR clarifies that `predict_single_npy_array` must use SimpleITK axis ordering. _(For other prediction functions, nnUNet's I/O functionality will do the axis re-ordering automatically. But, for `predict_single_npy_array`, axis re-ordering must be done by the caller before passing the array.)_

I've tried to align the docstring of `predict_single_npy_array` with the documentation in `inference/readme.md`, that way if someone jumps straight to the source code (like I did), they won't miss any important info.

I've also added some example code for `nibabel` users that clarifies the necessary steps to use `predict_single_npy_array`, since otherwise we had to figure that out on our own. _(But, perhaps this shouldn't be added, since nnUNet probably wants to encourage use of its own IO functions? Lmk and I can remove that addition if necessary.)

## Related issues/PRs

Fixes https://github.com/MIC-DKFZ/nnUNet/issues/1951.

### DIFF
--- a/nnunetv2/inference/predict_from_raw_data.py
+++ b/nnunetv2/inference/predict_from_raw_data.py
@@ -422,6 +422,9 @@ class nnUNetPredictor(object):
                                  output_file_truncated: str = None,
                                  save_or_return_probabilities: bool = False):
         """
+        input_image must use SimpleITK axis ordering!
+            (NB: if array comes from a nibabel-loaded image, you must swap the
+             axes of both the array *and* the spacing from [x,y,z] to [z,y,x]!)
         image_properties must only have a 'spacing' key!
         """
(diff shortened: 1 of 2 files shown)


FabianIsensee:
Thanks, I extended your comments because, really, you should never use plain nibabel to load images. Always use one of the nnU-Net imageio classes (NibabelIO). Even better, use the same as was used in preprocessing. You can check which one that was in the plans file.
There can even be conflicts between nnU-Net io classes as `NibabelIO` and `NibabelIOWithReorient` return different axes orderings by design. So really, just use the same as was used by nnU-Net

PR #2046, opened 2024-03-28
Improving planning and preprocessing

ancestor-mithril:
* Replacing multiple calls to `np.percentile` with one by passing all the percentiles to be calculated.
* Using `np.minimum` where possible to calculate the element-wise minimum of two arrays.


FabianIsensee:
Thanks a lot!

PR #2049, opened 2024-03-28
Improved crop_to_nonzero

ancestor-mithril:
The code path used during inference avoids multiple memory allocations which are costly for large volumes.


FabianIsensee:
Thanks!

PR #2079, opened 2024-04-11
Improving image io

ancestor-mithril:
* reducing the number of allocations when reading images and writing segmentations
* segmentation are casted to np.uint8 only if they are not already np.uint8
* images are stacked directly into a float32 array, without additional casts

### DIFF
--- a/nnunetv2/imageio/natural_image_reader_writer.py
+++ b/nnunetv2/imageio/natural_image_reader_writer.py
@@ -56,13 +56,13 @@ class NaturalImage2DIO(BaseReaderWriter):
             print('Image files:')
             print(image_fnames)
             raise RuntimeError()
-        return np.vstack(images).astype(np.float32), {'spacing': (999, 1, 1)}
+        return np.vstack(images, dtype=np.float32, casting='unsafe'), {'spacing': (999, 1, 1)}
 
     def read_seg(self, seg_fname: str) -> Tuple[np.ndarray, dict]:
         return self.read_images((seg_fname, ))
 
     def write_seg(self, seg: np.ndarray, output_fname: str, properties: dict) -> None:
-        io.imsave(output_fname, seg[0].astype(np.uint8), check_contrast=False)
+        io.imsave(output_fname, seg[0].astype(np.uint8, copy=False), check_contrast=False)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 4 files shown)


nnunetv2/imageio/natural_image_reader_writer.py
```diff
@@ -56,13 +56,13 @@ def read_images(self, image_fnames: Union[List[str], Tuple[str, ...]]) -> Tuple[
             print('Image files:')
             print(image_fnames)
             raise RuntimeError()
-        return np.vstack(images).astype(np.float32), {'spacing': (999, 1, 1)}
+        return np.vstack(images, dtype=np.float32, casting='unsafe'), {'spacing': (999, 1, 1)}
```
FabianIsensee, line 59: Can you comment on the use of 'unsafe', any specific reason this is picked?

nnunetv2/imageio/nibabel_reader_writer.py
```diff
@@ -78,14 +78,13 @@ def read_images(self, image_fnames: Union[List[str], Tuple[str, ...]]) -> Tuple[
...
 
-        stacked_images = np.vstack(images)
         dict = {
             'nibabel_stuff': {
                 'original_affine': original_affines[0],
             },
             'spacing': spacings_for_nnunet[0]
         }
-        return stacked_images.astype(np.float32), dict
+        return np.vstack(images, dtype=np.float32, casting='unsafe'), dict
```
FabianIsensee, line 87: same question as above

nnunetv2/imageio/natural_image_reader_writer.py
ancestor-mithril: I've chosen 'unsafe' for backwards compatibility as `astype` uses 'unsafe' casting (https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html). 
Moreover, 'unsafe' casting is needed in this case because numpy raises an exception when casting from int32 to float32:
```py
>>> import numpy as np
>>> x = np.array([1, 2, 3], dtype=np.int32)
>>> x.astype(np.float32)
array([1., 2., 3.], dtype=float32)
>>> x.astype(np.float32, casting='safe') 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Cannot cast array data from dtype('int32') to dtype('float32') according to the rule 'safe'
```

nnunetv2/imageio/nibabel_reader_writer.py
ancestor-mithril: Casting from integers to floating point can only be done using 'unsafe' casting.

FabianIsensee:
Thanks!

PR #2129, opened 2024-04-24
imported Union in run_case_npy

sarbabi:
To address NameError: name 'Union' is not defined


FabianIsensee:
Also see #2117. Thanks a lot for helping!

PR #2131, opened 2024-04-24
Update default_preprocessor.py

rohand24:
Union not imported from typing module


FabianIsensee:
Also see #2117. Thanks a lot for helping!

PR #2147, opened 2024-04-28
feat: BraTS dataset Conversion

strasserpatrick:
I think it was already in the codebase for nnunetv1, but the old script is no longer compatible. These two are runnable now with v2 dataset format


FabianIsensee:
Thanks!

PR #2148, opened 2024-04-30
[Sleep] SleepReaderWriter for nnUNet 

TomMingasson:


FabianIsensee:
Hey, this seems like a highly specific reader/writer. Do you see a larger community that would use & need this?
One big problem here is that you are associating a very generic file ending (.npy) with a highly specialized use case. This causes conflicts because in the absence of a specifically defined reader/writer we will infer one based on the supported file endings. This could cause problems with other generic npy-based datasets

TomMingasson:
> Hey, this seems like a highly specific reader/writer. Do you see a larger community that would use & need this? One big problem here is that you are associating a very generic file ending (.npy) with a highly specialized use case. This causes conflicts because in the absence of a specifically defined reader/writer we will infer one based on the supported file endings. This could cause problems with other generic npy-based datasets

Hi, 
I actually did not want to merge anything to `master`. 
It is indeed very specific to my problem ..  I just forked the project at the end.
Sorry about that.
Thanks !

PR #2154, opened 2024-05-02
add_max_num_epochs

mipbkhn:
Add an option to set the number of epochs to enable quick test trainings or early stop the training for applications that would take too much time to run 1000 epochs (as now done by default in the codes) and still not converge.


FabianIsensee:
Hi, this intentionally does not exist because it interferes with nnU-nets saving infrastructure. This is why I would like to avoid having that feature in the master. Please create separate trainers for shorter trainings or early stopping

PR #2158, opened 2024-05-03
Fixing dimension order: natural_image_reader_writer.py

felixkrones:
Fixing that additional dimension is added in front, such that new shape is (1,c,X,Y). If not, this is causing an "Error: Unexpected number of modalities" during verify for RGB images.


FabianIsensee:
Hey this fix is not correct. My comment in the file was wrong: the expected channel ordering is (c, 1, x, y) and that is what my current implementation does. Images will be stacked in the first dimension. The 1 is just to make this a pseudo 3d image in order to keep the dimensions within nnU-Net consistent

PR #2249, opened 2024-05-31
Fix documentation of nibabelIO's supported file types

mzenk:
With the `NibabelIO` rw classes, .mha and .nrrd files currently cause errors, although they are listed as supported file endings. It seems like nibabel does currently not natively support either of them. [Open issue for nibabel and nrrd](https://github.com/nipy/nibabel/issues/356)

Changes: Remove `.mha` and `.nrrd` from the supported file endings.


FabianIsensee:
Thanks Max! <3

PR #2269, opened 2024-06-07
fix: make sure that a custom output folder for cv results is empty

mzenk:
Issue:
When using `nnUNetv2_accumulate_crossval_results` and specifying an _existing_ output folder with argument `-o`, the specified folder is first deleted completely and re-created before CV results are saved in it. This is unexpected behavior and might need to undesired data loss.

Fix:
Require the directory specified with `-o` to be non-existent or empty.

Comment on alternative solution:
Another option would be to remove the `shutil.rmtree` in `nnunetv2/evaluation/accumulate_cv_results.py`, but this would affect the case where `-o` is not specified in the `nnUNetv2_accumulate_crossval_results` entrypoint. To keep the changes minimal, the `rmtree` command is kept as is.

### DIFF
--- a/nnunetv2/evaluation/find_best_configuration.py
+++ b/nnunetv2/evaluation/find_best_configuration.py
@@ -3,8 +3,9 @@ import os.path
 from copy import deepcopy
 from typing import Union, List, Tuple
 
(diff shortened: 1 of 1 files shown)


FabianIsensee:
Good find, much appreciated!

PR #2283, opened 2024-06-11
add optional arg to specify number of epochs

andylitalo:
Adds an optional argument to allow user to specify the number of epochs to train. Useful when the number of epochs desired has not been specified in one of the available trainers. If not provided, uses default from the specified trainer.

### DIFF
--- a/nnunetv2/run/run_training.py
+++ b/nnunetv2/run/run_training.py
@@ -35,7 +35,8 @@ def get_trainer_from_args(dataset_name_or_id: Union[int, str],
                           trainer_name: str = 'nnUNetTrainer',
                           plans_identifier: str = 'nnUNetPlans',
                           use_compressed: bool = False,
-                          device: torch.device = torch.device('cuda')):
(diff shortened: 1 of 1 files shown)


FabianIsensee:
Id like to kepp this as it currently is because this will interfere with nnU-Net's saving infrastructure and cause reproducibility problems, also see https://github.com/MIC-DKFZ/nnUNet/pull/2154

andylitalo:
Thanks for the response Fabian! And thanks for explaining the bug this change would introduce. My apologies for the replicate PR.

The trainers provided are helpful, but would it be possible to add one that trains for 500 epochs and another for 750? The gap between 250 and 1000 is large and sometimes includes the desired stopping time. I know developers can create such trainers on local forks, but sometimes it is convenient to install nnU-Net over PyPI instead of `git` (e.g., when installing on a docker image).

FabianIsensee:
I see. I have added the requested trainers

PR #2293, opened 2024-06-14
Fix #2292

### LINKED ISSUE #2292: Fix folder name for validation outputs in "How to run nnU-Net on a new dataset" documentation page
The folder in which the prediction output from validation cases are stored is named "validation". However, in the documentation, it is referred to as "validation_raw".

santhisenan:
The folder in which the prediction output from validation cases are stored is named "validation". However, in the documentation, it is mentioned as "validation_raw".

Corresponding issue: https://github.com/MIC-DKFZ/nnUNet/issues/2292


FabianIsensee:
Thanks!

PR #2303, opened 2024-06-18
Added Hausdorff Distance's 95th percentile

Zhack47:


FabianIsensee:
This PR is pretty bloated with a lot of stuff that does not match its title. Can you please create a more focused PR?
I am not sure if we want to have HD95 included in the default metrics tbh. The computation is rather slow and I would encourage users to compute metrics themselves using an external repository like medpy or the deepmind implementation if they need it.

PR #2313, opened 2024-06-23
Add multiprocessing for building paths list

hungvo304ml:
This helps improve the speed of building a list of paths during the initial stage of fingerprint extraction. Originally, the code was written in a nested loop fashion, which makes processing training datasets with large amounts of images significantly slow and waste a lot of time. This new version improves the speed by using multiprocessing.


FabianIsensee:
Thanks1

PR #2329, opened 2024-06-28
Fix np in preprocess

dani-capellan:
Fix number of processes in plan_and_preprocess_entrypoints.py - changed default to None and managed afterwords as in plan_and_preprocess_entry()


FabianIsensee:
THanks a lot!

PR #2356, opened 2024-07-08
Implement Argmax Inference Instead of Softmax when 'has_regions' is False

PengchengShi1220:
Issue: I have encountered a memory consumption issue in my Docker setup during inference tasks. Specifically, when the shm-size is set to 30GB, the system can successfully process softmax operations on input tensors of size [24, 724, 435, 435]. However, reducing the shm-size to 15GB leads to the unexpected termination of the Docker container due to insufficient shared memory during softmax computation.

Solution: Implement argmax instead of softmax when the 'has_regions' flag is set to False. This change is intended to address the shared memory (shm-size) constraints within Docker environments. It has been observed that the Docker container can handle the computation effectively, even with the shm-size set to 0, after making the proposed change.

### DIFF
--- a/nnunetv2/inference/export_prediction.py
+++ b/nnunetv2/inference/export_prediction.py
@@ -32,15 +32,22 @@ def convert_predicted_logits_to_segmentation_with_correct_shape(predicted_logits
                                             properties_dict['shape_after_cropping_and_before_resampling'],
                                             current_spacing,
                                             [properties_dict['spacing'][i] for i in plans_manager.transpose_forward])
-    # return value of resampling_fn_probabilities can be ndarray or Tensor but that does not matter because
-    # apply_inference_nonlin will convert to torch
-    predicted_probabilities = label_manager.apply_inference_nonlin(predicted_logits)
-    del predicted_logits
-    segmentation = label_manager.convert_probabilities_to_segmentation(predicted_probabilities)
+    if label_manager.has_regions:
+        # return value of resampling_fn_probabilities can be ndarray or Tensor but that does not matter because
+        # apply_inference_nonlin will convert to torch
+        predicted_probabilities = label_manager.apply_inference_nonlin(predicted_logits)
+        del predicted_logits
+        segmentation = label_manager.convert_probabilities_to_segmentation(predicted_probabilities)
 
-    # segmentation may be torch.Tensor but we continue with numpy
-    if isinstance(segmentation, torch.Tensor):
-        segmentation = segmentation.cpu().numpy()
+        # segmentation may be torch.Tensor but we continue with numpy
+        if isinstance(segmentation, torch.Tensor):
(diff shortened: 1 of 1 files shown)


FabianIsensee:
Thanks for this PR. I am not a big fan of this to be honest because we take something that is well structured and abstracted (labelmanager is in charge) and are changing it to be more complicated and moving functionality around. The original idea was that labelmanager is the central authority deciding how predictions should be handled, so that we only need to maintain and change the way things are handled in one location. If we now start re-adding label handling into different functions in different locations then it all becomes a complicated mess again. 
So I'd much rather better understand the underlying reason for this problem and address it systematically. Why does your implementation not cause as much SHM as mine. Does the overall amount of RAM needed to do the export change? Why not just give the docker more SHM?

PengchengShi1220:
Hi,
Thank you for your feedback. I encountered this issue while preparing for [AortaSeg24 Grand Challenge](https://aortaseg24.grand-challenge.org/). According to the Grand Challenge platform's requirements for algorithms ([Grand Challenge Documentation](https://grand-challenge.org/documentation/runtime-environment/)), 50% of the system memory is shared. For example, on a 16 GB instance, `/dev/shm` will be 8 GB, and this percentage is not modifiable. The provided instance is `ml.g4dn.2xlarge` with 1x NVIDIA T4 GPU, 8 vCPUs, 32 GB of memory, and 1 x 225 GB NVMe SSD. Therefore, participants can use a maximum memory size of 32 GB and an `shm-size` of 16 GB.

When I set up a new Docker environment, I observed a similar issue. In this Docker environment, setting the memory size to 35 GB and the `shm-size` to 35 GB allows the system to successfully process softmax operations on input tensors of size [24, 724, 435, 435]. However, setting the memory size to 30 GB and the `shm-size` to 30 GB leads to the unexpected termination of the Docker container. This suggests that inference on devices with smaller 30 GB memory sizes will encounter similar issues.

Regarding your point about label management, I agree that handling labels in one location is more efficient and avoids redundancy. However, I am currently unsure how to directly add an option for the inference step specifically within the label manager. I hope you can provide further suggestions on how to address this issue systematically.

FabianIsensee:
Hey, so I would like to close this PR because it breaks with our existing design patterns to work around a problem caused by a single challenge platform. I suggest users write their own inference code based on nnU-Net's functionality to work around these problems. You can take inspiration from our submission to ToothFairy2 which had way more of these problems because of the 43 labels this dataset has:
https://github.com/MIC-DKFZ/nnUNet/blob/master/documentation/competitions/Toothfairy2/inference_script_semseg_only_customInf2.py

Our script is a bit more involved than what you would need but it has some nice optimizations for runtime and memory efficiency

PengchengShi1220:
Hi, thanks for providing the detailed inference code for ToothFairy2. It has been really helpful for submitting inference code to the challenge platform. I will close this pull request.

PR #2384, opened 2024-07-22
Fix dataset order in move plans command

PaulBrachmann:
Since the target dataset has been planned before and its plans should be applied to the pretraining dataset, not the other way around, the two dataset arguments should be swapped.

E.g., the following yields the expected behavior with `30` being the pretraining and `31` the target dataset:
```sh
nnUNetv2_move_plans_between_datasets -s 31 -t 30 -sp nnUNetPlans -tp preTrainingPlans
```


FabianIsensee:
@seziegler can you take a look at this? Is this still an issue?

seziegler:
This pull request is based on an older commit. In the current version of the documentation this is fixed already and the current version is correct.

FabianIsensee:
Alright, thanks! And thanks @PaulBrachmann for your egnagement!

Issue #291, opened 2020-08-11
mmap length is greater than file size (error at beginning of training)


perecanals:
Hi Fabian,

I have been using nnunet for a while now and I came across error that is suddenly preventing me from performing training (I post all the printed verbose from the start of training for context):

```
Please cite the following paper when using nnUNet:
Fabian Isensee, Paul F. Jäger, Simon A. A. Kohl, Jens Petersen, Klaus H. Maier-Hein "Automated Design of Deep Learning Methods for Biomedical Image Segmentation" arXiv preprint arXiv:1904.08128 (2020).
If you have questions or suggestions, feel free to open an issue at https://github.com/MIC-DKFZ/nnUNet

###############################################
I am running the following nnUNet: 3d_lowres
My trainer class is:  <class 'nnunet.training.network_training.nnUNetTrainerV2.nnUNetTrainerV2'>
For that I will be using the following configuration:
num_classes:  1
modalities:  {0: 'CT'}
use_mask_for_norm OrderedDict([(0, False)])
keep_only_largest_region None
min_region_size_per_class None
min_size_per_class None
normalization_schemes OrderedDict([(0, 'CT')])
stages...

stage:  0
{'batch_size': 2, 'num_pool_per_axis': [4, 5, 4], 'patch_size': array([112, 192, 112]), 'median_patient_size_in_voxels': array([178, 300, 178]), 'current_spacing': array([1.23372493, 1.14846823, 1.23372493]), 'original_spacing': array([0.4296875, 0.3999939, 0.4296875]), 'do_dummy_2D_data_aug': False, 'pool_op_kernel_sizes': [[2, 2, 2], [2, 2, 2], [2, 2, 2], [2, 2, 2], [1, 2, 1]], 'conv_kernel_sizes': [[3, 3, 3], [3, 3, 3], [3, 3, 3], [3, 3, 3], [3, 3, 3], [3, 3, 3]]}

stage:  1
[...]

PR #2388, opened 2024-07-23
Fix for issue 291: mmap length is greater than file size

amithjkamath:
Dear @FabianIsensee  and others: It appears there are at least two of us who have been able to work around the issue in https://github.com/MIC-DKFZ/nnUNet/issues/291 with the simple fix of changing calls to:

np.load(... , 'r')

to 

np.load(... , 'r+')

This PR should ideally then not affect any other parts of the codebase, hence being safe to merge. Should you have any questions about this, please let me know! 

Amith


FabianIsensee:
This doesn't appear to be a sensible solution as this opens the file for writing as well. We don't want to do that because we only want to read the preprocessed data -> it's inherently an unsafe operation with potentially problematic consequences.
`mmap length is greater than file size` is an error that occurs when the writing of the memmap fails for some reason, be that insufficient storage or the operation being canceled mid-way. The recommended solution is to rerun the preprocessing

PR #2414, opened 2024-07-31
solving np.vstack numpy problem

victorcaquilpan:
I had a problem running the line 87 due to modifications in the behavior of the function np.vstack in Numpy.

Before:
return np.vstack(images, dtype=np.float32, casting='unsafe'), dict
Now:
return np.vstack(images).astype(np.float32, casting='unsafe'), dict

### DIFF
--- a/documentation/how_to_use_nnunet.md
+++ b/documentation/how_to_use_nnunet.md
@@ -71,7 +71,7 @@ nnUNetv2_train DATASET_NAME_OR_ID UNET_CONFIGURATION FOLD [additional options, s
 
 UNET_CONFIGURATION is a string that identifies the requested U-Net configuration (defaults: 2d, 3d_fullres, 3d_lowres, 
 3d_cascade_lowres). DATASET_NAME_OR_ID specifies what dataset should be trained on and FOLD specifies which fold of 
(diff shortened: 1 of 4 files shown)


FabianIsensee:
Hey can you please elaborate on what your problem with vstack is? We have not observed any problems here

victorcaquilpan:
Newer version of numpy don't allow defining the dtype inside the vstack method. Then, to replace `np.vstack(images, dtype=np.float32, casting='unsafe')`, I need to change to `np.vstack(images).astype(np.float32, casting='unsafe')`. I am using numpy 1.24.4

ancestor-mithril:
> Newer version of numpy don't allow defining the dtype inside the vstack method. Then, to replace `np.vstack(images, dtype=np.float32, casting='unsafe')`, I need to change to `np.vstack(images).astype(np.float32, casting='unsafe')`. I am using numpy 1.24.4

The dtype keyword was added in numpy 1.24: https://numpy.org/doc/1.24/reference/generated/numpy.vstack.html.
Older versions of numpy are already too old.

FabianIsensee:
In that case the proper solution would be to require numpy>=1.24

FabianIsensee:
I'll change that

PR #2462, opened 2024-08-22
Fix `deep_supervision` parameter doesn't work in `nnunetv2.utilities.get_network_from_plans`

AIboy996:
- Bug: 
`nnunetv2.utilities.get_network_from_plans` has an optional parameter: `deep_supervision`, but it doesn't work.

- Fix: 

```python
...
    if deep_supervision is not None and 'deep_supervision' not in arch_kwargs.keys():
        arch_kwargs['deep_supervision'] = deep_supervision
```

to 

```python
...
    if deep_supervision is not None and 'deep_supervision' not in architecture_kwargs.keys():
        architecture_kwargs['deep_supervision'] = deep_supervision

```

- An example for this function:
```python
if __name__ == "__main__":
    import torch

    model = get_network_from_plans(
        arch_class_name="dynamic_network_architectures.architectures.unet.ResidualEncoderUNet",
        arch_kwargs={
            "n_stages": 7,
            "features_per_stage": [32, 64, 128, 256, 512, 512, 512],
            "conv_op": "torch.nn.modules.conv.Conv2d",
            "kernel_sizes": [[3, 3], [3, 3], [3, 3], [3, 3], [3, 3], [3, 3], [3, 3]],
            "strides": [[1, 1], [2, 2], [2, 2], [2, 2], [2, 2], [2, 2], [2, 2]],
            "n_blocks_per_stage": [1, 3, 4, 6, 6, 6, 6],
            "n_conv_per_stage_decoder": [1, 1, 1, 1, 1, 1],
            "conv_bias": True,
            "norm_op": "torch.nn.modules.instancenorm.InstanceNorm2d",
            "norm_op_kwargs": {"eps": 1e-05, "affine": True},
            "dropout_op": None,
            "dropout_op_kwargs": None,
            "nonlin": "torch.nn.LeakyReLU",
            "nonlin_kwargs": {"inplace": True},
        },
        arch_kwargs_req_import=["conv_op", "norm_op", "dropout_op", "nonlin"],
        input_channels=1,
        output_channels=4,
        allow_init=True,
        deep_supervision=True,
    )
    data = torch.rand((8, 1, 256, 256))
    target = torch.rand(size=(8, 1, 256, 256))
    output = model(data)
    assert isinstance(output, list), "deep-supervisison doesn't work"
```

- And remove a redundant line in `nnunetv2.training.nnUNetTrainer.nnUNetTrainer.py`
```python
...
        # make sure deep supervision is on in the network
[...]

PR #2508, opened 2024-09-19
Add a toy example for `nnunetv2.utilities.get_network_from_plans`

AIboy996:
As mentioned in #2462.


FabianIsensee:
Here we go :-)

PR #2535, opened 2024-10-09
Fix typo in set_environment_variables.md

Kirscher:
Fix typo in command prompt variable syntax: use `echo %variable_name%` instead of `echo %[variable_name]%`


vishwa702 (approved):

FabianIsensee:
Thanks!

PR #2538, opened 2024-10-09
added a function to save heatmaps

hadign20:
Added a function to nnunetv2\inference\export_prediction.py to generate heatmaps (PNG files) of the probabilities for each class of segmentations. I can make it a flag if it is preferred.

### DIFF
--- a/nnunetv2/inference/export_prediction.py
+++ b/nnunetv2/inference/export_prediction.py
@@ -4,12 +4,54 @@ from typing import Union, List
 
 import numpy as np
 import torch
+import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


FabianIsensee:
You are changing the way the segmentation export works. It is actually more precise to resample the probabilities and then convert to the final segmentation. The behavior you see is intentional. 
If you need speed for resampling, please use planners like `nnUNetPlannerResEncL_torchres`. They replace the slow default resampling with torch.interpolate. This comes at the cost of a very small drop in segmentation performance, but should still be better than the suggested solution
Best,
Fabian

Issue #2540, opened 2024-10-10
export_prediction_from_logits takes a very long time to finish


pooya-mohammadi:
The prediction process takes about 10-20 seconds on my system: 4090GPU
`100%|██████████| 120/120 [00:16<00:00,  7.30it/s]`
But the this function export_prediction_from_logits is super slow on my system with 32 cores of CPU
`[INFO] Elapsed time for export_prediction_from_logits: 534.6257283687592`
I added a simple time tracker. 
What could have caused this?

The number of the classes is equal to the number of the classes of Totalsegmentator for whole all items. 
I run the Totalsegmentator on the same image and it takes less than 30 seconds to finish!

Can any one help me on this?

pooya-mohammadi:
I also saved the data before passing to `export_prediction_from_logits` and ran it in a separate process and the output took the same amount of time while only 10-20 percent of the CPU is occupied!

I noticed that Totalsegmentator uses the same function, however, they resize/resample the inputs before passing it to nnunet predictor.

And the reason nnunet takes very long is because it resizes each class label one by one in this code snippet:
```
for c in range(data.shape[0]):
    reshaped_final[c] = resize_fn(data[c], new_shape, order, **kwargs)
```

@FabianIsensee Do you have any plans to change this or use a method like Totalsegmentator?
Or is there anyway to skip this part? For segmentations with large number of classes this becomes very slow.

ancestor-mithril:
The reason TS2 is fast is because it uses NN interpolation (creating less smooth masks). nnUNet uses tricubic resampling (or 3-linear resampling if you configure  your training to use torch resampling).

pooya-mohammadi:
@ancestor-mithril Thanks for your response can you show me how to do it?
However this code snippet is too slow:
```
for c in range(data.shape[0]):
    reshaped_final[c] = resize_fn(data[c], new_shape, order, **kwargs)
```
Do you have any comments on this?

pooya-mohammadi:
One more thing, since TS2 does not save probabilities, there is no need to do the resize on all the classes separately. And it only does a single zoom on the output segmentation.

ancestor-mithril:
You can change the resampling by using a different experiment planner (see Documentation). For example, you can use https://github.com/MIC-DKFZ/nnUNet/blob/master/nnunetv2/experiment_planning/experiment_planners/resampling/resample_with_torch.py.

ancestor-mithril:
[...]

PR #2545, opened 2024-10-11
Fast inference for large files with several classes #2540

pooya-mohammadi:
When doing inference on large files with large number of classes the following code section blocks the code for up too 600 seconds:

```
            for c in range(data.shape[0]):
                reshaped_final[c] = resize_fn(data[c], new_shape, order, **kwargs)
```
in `preprocessing.resampling.default_resampling.py` and func: `resample_data_or_seg`.
If someone does not need the probabilities they can skip this section by only sending the segmentation array which is very fast.

Solving issue #2540

### DIFF
--- a/nnunetv2/inference/export_prediction.py
+++ b/nnunetv2/inference/export_prediction.py
@@ -1,11 +1,10 @@
-import os
-from copy import deepcopy
+import time
 from typing import Union, List
 
 import numpy as np
 import torch
 from acvl_utils.cropping_and_padding.bounding_boxes import bounding_box_to_slice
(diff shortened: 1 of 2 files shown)


pooya-mohammadi:
Or the depicted code snippet can be done in a multiprocessing manner.

FabianIsensee:
The correct way of speeding up segmentation export is to use resample_torch_fornnunet. You can read about how to enable that here: https://github.com/MIC-DKFZ/nnUNet/issues/2912
This PR seems to contain a collection of changes. Please create isolated PRs for each change you propose. Ideally open up an issue beforehand so that we can discuss :-)

Issue #2563, opened 2024-10-21
continue training from checkpoint_best.pth instead of checkpoint_latest.pth


romainVala:
Hi

I do not understand why you choose to continue training from the checkpoint_latest.pth instead of checkpoint_best.pth.
Checkpoint_latest.pth is saved every 50 epochs, so when we restart; we may loos up to 49 epochs.
In my case, I have a very long training epoch time (up to 2000 s sometimes) do loosing 50 epoch is then equivalent to loose 27 hours of computing ...

I find a way around, by just removing the checkpoint_latest.pth from the training log dir.
see  https://github.com/MIC-DKFZ/nnUNet/blob/520e749402795ad00f025c70430115c3b8c7f6ca/nnunetv2/run/run_training.py#L81  

It would be more efficient to compare the date (or directly the epoch) of both checkpoint_best.pth checkpoint_latest.pth and choose the more recent

But may be I miss something ?

kndahl:
+

romainVala:
Is there a way to change the parameter that define the save period to 50 epochs ? 
In my case of very long training I would like to make it every 10 epochs

gaojh135:
> Is there a way to change the parameter that define the save period to 50 epochs ?有没有办法将定义保存周期的参数更改为 50 epochs ？ In my case of very long training I would like to make it every 10 epochs在我训练时间很长的情况下，我想每 10 个 epoch 进行一次

https://github.com/MIC-DKFZ/nnUNet/blob/1e2aa0a1cc206ab4ac8347517d6325c629ee611f/nnunetv2/training/nnUNetTrainer/nnUNetTrainer.py#L188

gaojh135:
> Hi 你好
> 
[...]

PR #2564, opened 2024-10-22
Use latest checkpoint in training

kndahl:
Solution for issue: https://github.com/MIC-DKFZ/nnUNet/issues/2563

Instead of loading the latest checkpoint when the flag `--c` is handled, we will filter to get the most recent checkpoint. This could be the latest, best, or final checkpoint.


fitzjalen:
@FabianIsensee please review it. As for me I always have to rename checkpoint if I stopped it before.

FabianIsensee:
Any update on this?

Issue #2556, opened 2024-10-16
Optional Torch Multiprocessing in nnUNet for Improved Security and Compatibility


LennyN95:
Dear nnUNet team,

We are currently facing challenges when running nnUNet in Docker containers.
The requirement to use flags like `--ipc=host` or `--shm-size` for `torch.multiprocessing` (as reported and discussed in previous issues) makes it difficult to deploy models in our [mhub.ai](https://mhub.ai) platform. We hesitate to suggest the use of `--ipc=host`, which is a simple solution but removes security restrictions and should therefore be used with caution. On the other hand, manually specifying the `shm-size` means an additional burden and makes the inference even more complicated from MHub's point of view.

We currently have some contributions to our portfolio on hold to discuss this topic. In our particular case, the inference is executed sequentially and therefore does not require multiprocessing per se. We propose to make the use of `torch.multiprocessing` optional during inference.

We welcome any comments and an open discussion on this topic.

Thank you very much!
Leo.

surajpaib:
Hi Team,

I would also like to address this issue but from the perspective of user debugging. I think having an option to set `num_workers=0` in the inference dataloader would allow users to better debug error traces that occur.

I see a few issues that would probably benefit from this as well:
https://github.com/MIC-DKFZ/nnUNet/issues/2509
https://github.com/MIC-DKFZ/nnUNet/issues/2514
https://github.com/MIC-DKFZ/nnUNet/issues/2182

Looking forward to hear what you think

@Lars-Kraemer

TaWald:
Sounds like a reasonable proposal.
Currently, time is short on our side so if one of you, @surajpaib or @LennyN95, would be willing to create a PR with the feature in question we can certainly integrate this.

FabianIsensee:
Setting num_workers=0 will probably be more work than just offering a new member function of nnUNetPredictor that covers this. This function would then not be supported via command line but if you are packing nnU-Net in a Docker you might not need that anyways. Would this be a sensible solution to your problem?
Best,
Fabian

PereGinebra:
I've come along similar issues when trying to deploy nnUNet models in docker containers, and although my multiprocessing issues might come from somewhere else, I also agree that having a sequential predictor might simplify a lot of things, especially when you don't need to do inference in big batches.

[...]

PR #2614, opened 2024-11-22
Optional Torch Multiprocessing in nnUNet for Improved Security and Compatibility

LennyN95:
This PR resolves https://github.com/MIC-DKFZ/nnUNet/issues/2556.

- Implementation of preprocessing based on a single process
- Introduce new environment variables to set `-npp` and `-nps`

We have done some internal testing and get _identical results_ between the latest `nnunetv2==2.5.1` and with our proposed patch.

### DIFF
--- a/nnunetv2/inference/data_iterators.py
+++ b/nnunetv2/inference/data_iterators.py
@@ -65,21 +65,35 @@ def preprocess_fromfiles_noqueue(list_of_lists: List[List[str]],
                                 dataset_json: dict,
                                 configuration_manager: ConfigurationManager,
                                 verbose: bool = False):
+    
     print("Running preprocessing in non-multiprocessing mode")
+    
     data_iterator = []
-    for i in range(len(list_of_lists)):
-        input_files = list_of_lists[i]
-        seg_file = list_of_segs_from_prev_stage_files[i] if list_of_segs_from_prev_stage_files is not None else None
-        output_file = output_filenames_truncated[i] if output_filenames_truncated is not None else None
+    label_manager = plans_manager.get_label_manager(dataset_json)
+    
+    for idx in range(len(list_of_lists)):
+        
+        input_files = list_of_lists[idx]
+        seg_file = list_of_segs_from_prev_stage_files[idx] if list_of_segs_from_prev_stage_files is not None else None
+        output_file = output_filenames_truncated[idx] if output_filenames_truncated is not None else None
+        
         preprocessor = configuration_manager.preprocessor_class(verbose=verbose)
         data, seg, data_properties = preprocessor.run_case(input_files, seg_file, plans_manager, configuration_manager, dataset_json)
+        
+        if list_of_segs_from_prev_stage_files is not None and list_of_segs_from_prev_stage_files[idx] is not None:
+                seg_onehot = convert_labelmap_to_one_hot(seg[0], label_manager.foreground_labels, data.dtype)
+                data = np.vstack((data, seg_onehot))
+        
         data = torch.from_numpy(data).to(dtype=torch.float32, memory_format=torch.contiguous_format)
+        
         preprocessed_data = {
             'data': data,
             'data_properties': data_properties,
             'ofile': output_file
         }
+        
         data_iterator.append(preprocessed_data)
+    
     return data_iterator
 
 def preprocessing_iterator_fromfiles(list_of_lists: List[List[str]],
@@ -91,6 +105,9 @@ def preprocessing_iterator_fromfiles(list_of_lists: List[List[str]],
                                      num_processes: int,
                                      pin_memory: bool = False,
                                      verbose: bool = False):
+    
+    num_processes = 1
+    
     if num_processes > 1:
         context = multiprocessing.get_context('spawn')
         manager = Manager()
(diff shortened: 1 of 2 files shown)


LennyN95:
@FabianIsensee I have noticed some (significant) differences between `nnunetv2==2.0` and `nnunetv2==2.5.1`. This is beyond the scope of this topic / PR, but since some of our contributors have used this version, we'd like to see if and to what extent we can offer them a solution as well. In general, I am curious if you have any idea what changes might have led to these differences (we are talking about a _Dice score of 0.9451_ between these versions for a lung nodule segmentation task).

FabianIsensee:
Hey, thanks for the PR and sorry for being slow. I have too much on my plate.
One thing I am not particularly fond of is that fact that your no queue functions preprocess all the data, store it in a list and then yield from the existing list. This causes a lot of unnecessary RAM consumption. Why not yield the items as they are ready? Tat way we don't have to keep them in memory.

Can you please provide more information on the inconsistency in performance? What is the difference between the runs? This doesn't become quite clear from your message

LennyN95:
Hey @FabianIsensee thanks for the reply and no worries!

> One thing I am not particularly fond of is that fact that your no queue functions preprocess all the data, store it in a list and then yield from the existing list. This causes a lot of unnecessary RAM consumption. Why not yield the items as they are ready? Tat way we don't have to keep them in memory.

Good point! @surajpaib Looks like we can combine it to output each item right after preprocessing.

> Can you please provide more information on the inconsistency in performance? What is the difference between the runs? This doesn't become quite clear from your message

I ran some tests (with [this submission model](https://github.com/MHubAI/models/pull/92)). All MHub models come with test and reference data, so I manually updated the version from `nnunetv2==2.0` to `nnunetv2==2.5.1` and compared the generated segmentation with the reference. Normally, we would expect a Dice score of ~1 (with slight variations due to rounding errors caused by different graphics card architectures). However, in this case I got a Dice score of `0.9451`, so the generated masks differ when using the latest version. I was wondering if you could link this to a specific change.

FabianIsensee:
Is there a way I can reproduce this locally to investigate? Like can you share both checkpoints + the reference data that gives Dice=1 in one case and 0.94 in the other? That way I can track down where things diverge

LennyN95:
Thank you @FabianIsensee for looking into this. 

You can use the **BAMF NNUnet Lung and Nodules V2** (https://github.com/MHubAI/models/pull/92) model for testing.
- The weights are available for download [here](https://zenodo.org/record/11582738/files/Dataset007_Nodules.zip).
- The sample input data and reference output can be downloaded [here](https://zenodo.org/records/13880663/files/bamf_nnunet_ct_lungnodules.test.zip?download=1).

You can also build and run the model via MHub in a self-contained environment by following these steps below: 

```bash
$LOCAL_NNUNET_PATCH_DIR=/absolute/path/to/loacal/nnunet/patch

# build the model container
docker build \
    -t mhubai-nnunet-test/bamf_nnunet_ct_lungnodules:latest \
    --build-arg MHUB_MODELS_REPO=https://github.com/bamf-health/mhub-models.git::bamf_nnunet_ct_lung_v2 \
    https://github.com/bamf-health/mhub-models.git#bamf_nnunet_ct_lung_v2:models/bamf_nnunet_ct_lungnodules/dockerfiles

# run the model container
docker run --rm -it --entrypoint bash --gpus all -v $LOCAL_NNUNET_PATCH_DIR:/nnunet-src mhubai-test/bamf_nnunet_ct_lungnodules:latest

# install nnunet in the container
uv pip install -e /nnunet-src

# update NNUnetRunnerV2 Module
sed -i 's/bash_command += \["-c", self.nnunet_config\]/bash_command += \["-c", self.nnunet_config, "-npp", "0", "-nps", "0"\]/' /app/models/bamf_nnunet_ct_lungnodules/utils/NNUnetRunnerV2.py

# run mhub test
mhub.test srmteyvx
```

Let me know if you need anything else or if I can assist you in any way!

FabianIsensee:
Hey, thanks for sharing. I will try to find time this week to look into this. Since I will be running this locally (no docker): All I need to do is run the prediction on the provided reference sample with both versions and compare?

LennyN95:
Hi @FabianIsensee, thank you!

> All I need to do is run the prediction on the provided reference sample with both versions and compare?

Correct!

I'm curios what you will find. Let me know if there is anything I can help with!

FabianIsensee:
Hey so I looked into this. Yes the segmentations generated by the two versions diifer. Here are the predictions I generated:
[testimage_v20.nii.gz](https://github.com/user-attachments/files/18654488/testimage_v20.nii.gz)
[testimagev252.nii.gz](https://github.com/user-attachments/files/18654490/testimagev252.nii.gz)
The difference is to be expected because the inference pipeline was rebuilt in the meantime and comes with a few improvements. These are mostly quality of life, but some also affect the predictions. 
When we made those changes we extensively evaluated that they would not result in a measurable performance difference. Specifically we reran the validations of our models and confirmed that the dice scores were comparable to the ones generated with the old setup. So the new results are different, but equivalently good. Have you tried running the validations of the 5 fold cross-validation with v2.0 and v2.5.2 and compared the dice scores? If you observe a substantial difference in this setup that would be very interesting and require more investigations on my end.

LennyN95:
@FabianIsenseeThank you for looking into this. This aligns with my findings. Upon inspection, the differences are limited to border voxels. The relatively large Dice difference can be attributed to the small size of the segmentations.

To move on, I created a new PR (#2705) using the proposed `predict_from_files_sequential()` function.

PR #2630, opened 2024-12-02
adding focal loss, distance map dice loss, loading distance maps

andy-s-ding:
Added a FocalLoss class that also works in compound loss functions. nnUNet also currently does not have a framework for loading in distance maps that were calculated offline. I propose a framework of calculating distance maps from segmentations in the `preprocessed` folder designated by the `plans` json. These maps are saved as `<case_num>_dist.npy`. In doing so, the `plans` json does not need to necessarily be modified. I have also made a Dataset class and a DataLoader class that is able to load these distance maps during training. Furthermore, I have added custom transform classes that transform distance maps commensurately with image data and segmentations. To make this work, I submitted a pull request in batchgeneratorsv2 to modify the BasicTransform class slightly.

### DIFF
--- a/nnunetv2/evaluation/evaluate_predictions.py
+++ b/nnunetv2/evaluation/evaluate_predictions.py
@@ -251,12 +251,13 @@ def evaluate_simple_entry_point():
 
 
 if __name__ == '__main__':
-    folder_ref = '/media/fabian/data/nnUNet_raw/Dataset004_Hippocampus/labelsTr'
-    folder_pred = '/home/fabian/results/nnUNet_remake/Dataset004_Hippocampus/nnUNetModule__nnUNetPlans__3d_fullres/fold_0/validation'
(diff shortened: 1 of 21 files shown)


FabianIsensee:
Hey,
thanks a lot for the effort and for putting this together. The FocalLoss addition is definitely useful — would be great to have that as a standalone PR so we can merge it cleanly.

As for the distance map pipeline: I totally get the motivation, and it’s cool to see you're already training with it. That said, it introduces quite a bit of additional complexity across preprocessing, transforms, and data loading — and without a clear or widely requested use case, I’m hesitant to bring that into the core codebase. The current pipelines are already quite complex, and maintaining forward compatibility (plus answering future issues) is not something I want to commit to unless the benefit is substantial.

So my suggestion would be: let’s keep the distance map logic as a custom extension for now, outside the main repo. If there’s more traction or a clear need in the future, we can definitely revisit it.
Best,
Fabian

PR #2647, opened 2024-12-17
Add basic GitHub action for testing nnunet inference

wasserth:
This adds a github action to automatically run a nnunet inference on a small CT example dataset for ubuntu + mac + windows. This is good to see if any new commit breaks parts of the inference on any operating system. Currently it fails for windows. Therefore windows is commented out at the moment. Whenever the windows problem is fixed or if you want to debug it, you can simply activate windows in the `run_tests_nnunet.yml` file.


.gitignore
vishwa702, line 1: Use `!` to negate some example files if needed. See [exceptions in gitignore](https://stackoverflow.com/questions/2415873/exceptions-in-gitignore).

FabianIsensee:
Thanks Jakob! :heart_hands:

PR #2676, opened 2025-01-27
Feature/bottleneck embeddings

eliotbehr:


FabianIsensee:
Hey, thanks for sharing this! I would however prefer not to merge it as the code makes hard assumptions about the network architecture that is being used and will be incompatible with custom architectures. It also expects to fit the entire image in the network at once. This is often not the case (stride induced image shape constraint, VRAM limitations, disparity between image and patch size) and would fail for too many users (or produce meaningless results). So while this works for your application specifically it is not something that can just be let out in the wild. We would have to do a lot of maintenance and explaining in the github issues

Issue #2681, opened 2025-01-28
PyTorch 2.6 has changed the default value of `weights_only` to `True`, breaking the loading of some `nnunetv2` models


joshuacwnewton:
> Upstream context: https://github.com/pytorch/pytorch/issues/52181, https://github.com/pytorch/pytorch/pull/137602

PyTorch 2.6 isn't out on PyPI yet, but it just dropped on the http://download.pytorch.org/whl/cpu/torch/, and so we ran into the following class of error:

```
_pickle.UnpicklingError: Weights only load failed. This file can still be loaded, to do so you have two options, do those steps only if you trust the source of the checkpoint. 
	(1) In PyTorch 2.6, we changed the default value of the `weights_only` argument in `torch.load` from `False` to `True`. Re-running `torch.load` with `weights_only` set to `False` will likely succeed, but it can result in arbitrary code execution. Do it only if you got the file from a trusted source.
	(2) Alternatively, to load with `weights_only=True` please check the recommended steps in the following error message.
	WeightsUnpickler error: Unsupported global: GLOBAL torch.optim.adam.Adam was not an allowed global by default. Please use `torch.serialization.add_safe_globals([Adam])` or the `torch.serialization.safe_globals([Adam])` context manager to allowlist this global if you trust this class/function.
```

Full list of errors (and the relevant globals) below:

<details>

```
 =========================== short test summary info ============================
FAILED testing/cli/test_sct_deepseg.py::test_segment_nifti_binary_seg[/__w/spinalcordtoolbox/spinalcordtoolbox/data/sct_testing_data/t2s/t2s.nii.gz-/__w/spinalcordtoolbox/spinalcordtoolbox/data/sct_testing_data/t2s/t2s_seg-deepseg.nii.gz-t2s_seg_deepseg.nii.gz-seg_sc_t2star-0.9] - _pickle.UnpicklingError: Weights only load failed. This file can still be loaded, to do so you have two options, do those steps only if you trust the source of the checkpoint. 
	(1) In PyTorch 2.6, we changed the default value of the `weights_only` argument in `torch.load` from `False` to `True`. Re-running `torch.load` with `weights_only` set to `False` will likely succeed, but it can result in arbitrary code execution. Do it only if you got the file from a trusted source.
	(2) Alternatively, to load with `weights_only=True` please check the recommended steps in the following error message.
[...]

PR #2683, opened 2025-01-30
set weights_only=False in torch.load

wasserth:
Fixes this critical issue: https://github.com/MIC-DKFZ/nnUNet/issues/2681

### DIFF
--- a/nnunetv2/inference/predict_from_raw_data.py
+++ b/nnunetv2/inference/predict_from_raw_data.py
@@ -82,7 +82,7 @@ class nnUNetPredictor(object):
         for i, f in enumerate(use_folds):
             f = int(f) if f != 'all' else f
             checkpoint = torch.load(join(model_training_output_dir, f'fold_{f}', checkpoint_name),
-                                    map_location=torch.device('cpu'))
+                                    map_location=torch.device('cpu'), weights_only=False)
             if i == 0:
                 trainer_name = checkpoint['trainer_name']
                 configuration_name = checkpoint['init_args']['configuration']


snorthman:
Rather than the suggested fix here,

```            
checkpoint = torch.load(join(model_training_output_dir, f'fold_{f}', checkpoint_name),
                                    map_location=torch.device('cpu'))
                                    map_location=torch.device('cpu'), weights_only=False)
```

may I instead suggest

```
import _codecs
torch.serialization.add_safe_globals([np.core.multiarray.scalar, np.dtype, np.dtypes.Float64DType, _codecs.encode])
checkpoint = torch.load(join(model_training_output_dir, f'fold_{f}', checkpoint_name),
                                    map_location=torch.device('cpu'))
```

FabianIsensee:
Hey @snorthman can you please elaborate on why your suggestion is to be preferred? I would like to understand the differences

FabianIsensee:
I am merging @wasserth PR for now since this seems to be the most straightforward (+ it's urgent) but I am open for your variant as well

snorthman:
My fix respects the warnings posed by torch: 
`"Re-running torch.load with weights_only set to False will likely succeed, but it can result in arbitrary code execution. Do it only if you got the file from a trusted source."`

When torch loads a file, it only loads types on the "safe globals" list, which initially doesn't include some types used in nnUNet, most of which are numpy types. The end results is of course the same, whether we use `weight_only=False` or my solution, but my solution respects the warnings posed by torch.

I should note that the list of globals I put forward is what worked for me, but perhaps using different settings there may be more types needed for this to be a catch-all solution.

jamesobutler:
@snorthman Have you issued a PR to this repo with your proposal?

PR #2684, opened 2025-01-30
Change torch.cuda.amp.GradScaler() to torch.GradScaler("cuda") 

V-kr0pt:
"torch.cuda.amp.GradScaler(args...) will be deprecated using torch.GradScaler("cuda", args...) instead "

[pytorch documentation](https://pytorch.org/docs/stable/amp.html)


FabianIsensee:
thank you!

PR #2692, opened 2025-02-05
Flexible number of epochs

lukas-folle-snkeos:
Allows to limit the number of epoch over the os environment variable `num_epochs`.


FabianIsensee:
This will cause chaos because nnU-Net saves by trainer name and one will no longer be able to see what experiment was run with how many epochs. I would like to keep this explicit as we do it for example here:
https://github.com/MIC-DKFZ/nnUNet/blob/master/nnunetv2/training/nnUNetTrainer/variants/training_length/nnUNetTrainer_Xepochs.py

PR #2728, opened 2025-03-04
Bugfix - predict_from_data_iterator tensor serialization issue

Kenneth-Schroeder:
I have encountered several fatal SIGBUS errors using TotalSegmentator which uses nnUnetV2 under the hood. 
The errors occurred when running TotalSegmentator on large inputs.
After some digging, I was able to boil the issue down to the predict_from_data_iterator function of nnUNet, which passes the (in my case quite large) prediction result tensors to worker processes. For this step, the tensors need to be serialized, which apparently can cause memory alignment issues and trigger SIGBUS errors.
I noticed the error would disappear if I removed the multiprocessing logic, but numpy arrays seem much more stable regarding serialization and fixed the SIGBUS errors even with multiprocessing, hence this PR.

Here's a summary from Claude AI with slightly more details:

# SIGBUS Error in PyTorch Multiprocessing: Claude (AI) Summary

The SIGBUS error occurred when passing PyTorch tensors between processes via multiprocessing. Here's what happened:

## Root Cause

In the original function, we were serializing PyTorch tensors directly for multiprocessing:

```python
# This line produces a PyTorch tensor
prediction = self.predict_logits_from_preprocessed_data(data).cpu()

# Problematic: Passing PyTorch tensor directly to worker process
future = executor.submit(
    export_prediction_from_logits,
    prediction, properties, self.configuration_manager, 
    # ...other arguments...
)
```

When PyTorch tensors are serialized and sent to child processes, their complex memory layout isn't perfectly preserved. This leads to memory alignment issues that trigger SIGBUS errors.

## Solution

Converting the tensor to a NumPy array before passing it to the worker process:

```python
# Convert to NumPy for safer serialization
prediction = self.predict_logits_from_preprocessed_data(data).cpu().numpy()

# Now safe: Passing NumPy array to worker process
future = executor.submit(
    export_prediction_from_logits,
    prediction, properties, self.configuration_manager, 
    # ...other arguments...
)
```

This fixes the issue by:
- Converting the tensor to a simpler NumPy array format
- Creating a copy with proper memory alignment
- Eliminating PyTorch-specific memory structures that don't serialize well

The SIGBUS error only happens with multiprocessing because tensors must cross process boundaries through serialization/deserialization, which can't perfectly preserve their memory layout.

### DIFF
--- a/nnunetv2/inference/predict_from_raw_data.py
+++ b/nnunetv2/inference/predict_from_raw_data.py
@@ -385,7 +385,7 @@ class nnUNetPredictor(object):
                     proceed = not check_workers_alive_and_busy(export_pool, worker_list, r, allowed_num_queued=2)
 
                 # convert to numpy to prevent uncatchable memory alignment errors from multiprocessing serialization of torch tensors
-                prediction = self.predict_logits_from_preprocessed_data(data).cpu().detach().numpy().copy()
+                prediction = self.predict_logits_from_preprocessed_data(data).cpu().detach().numpy()
 
                 if ofile is not None:
                     print('sending off prediction to background worker for resampling and export')
--- a/nnunetv2/inference/predict_from_raw_data.py
+++ b/nnunetv2/inference/predict_from_raw_data.py
@@ -385,7 +385,7 @@ class nnUNetPredictor(object):
                     proceed = not check_workers_alive_and_busy(export_pool, worker_list, r, allowed_num_queued=2)
 
                 # convert to numpy to prevent uncatchable memory alignment errors from multiprocessing serialization of torch tensors
-                prediction = self.predict_logits_from_preprocessed_data(data).cpu().detach().numpy()
+                prediction = self.predict_logits_from_preprocessed_data(data).cpu().detach().numpy().copy()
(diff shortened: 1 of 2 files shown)


Kenneth-Schroeder:
Hi @FabianIsensee, thanks for self-assigning.
This is a high-priority PR for us and should be quite simple to review.
If you could find some time to look at it soon, that would be great.

ValentinKoch:
Any chance of merging this soonish @FabianIsensee ?

FabianIsensee:
Hey Kenneth, thanks for the PR! And sorry for being late. I have too many responsibilities and can impossibly keep up with everything.
I merged your changes, but am not happy with changing from tensor to array. We communicate tensors for a reason. They can be transferred between processes MUCH faster than arrays can. (Not familiar enough with the details to know why that is)
To me this is more of a workaround and not a final solutions. It seems like a thing pytorch should/chould fix on their end. Have you opened an issue on their end about this?
Since I don't have time to do extensive testing at the moment I am holding off with pushing a new release to pypi. Do you need this urgently?
Best,
Fabian

PR #2748, opened 2025-03-21
Remove dicom2nifti since it isn't used

### LINKED ISSUE #2430: Fatal FIPS Failure
My work is interested in trying nnUNet models and seeing how they perform for what we require. We are required to be FIPS compliant. 
I can not get past the plan_and_preprocess command without recieving the FATAL FIPS SELFTEST FAILURE. I have attempted to use v1 and v2 for this and run into the same issue.  One of the thoughts is the issue has to do with the just-in-time compiler using md5 hash which is not FIPS compliant. 

I have attempted to follow this v1 example:
https://github.com/MIC-DKFZ/nnUNet/blob/nnunetv1/documentation/training_example_Hippocampus.md

I have been following the documentation for v2 using the same Hippocampus dataset.

Is there a way around this? Is there anything else you require from me? 

Thanks for your assistance. It is greatly appreciated.

Edit: I've given myself a sterile python environment. Updated photo as the deprecation warnings are not related.
Code_hsa0nzS10j

MattTheCuber:
Closes #2430

I can't find anywhere that uses the `dicom2nifti` library and it is causing a fatal FIPS error on FIPS enabled machines. Removing the dependency would resolve the issue.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,6 @@ dependencies = [
     "acvl-utils>=0.2.3,<0.3",  # 0.3 may bring breaking changes. Careful!
     "dynamic-network-architectures>=0.3.1,<0.4",  # 0.3.1 and lower are supported, 0.4 may have breaking changes. Let's be careful here
     "tqdm",
-    "dicom2nifti",
     "scipy",
     "batchgenerators>=0.25.1",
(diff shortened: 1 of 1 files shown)


vmiller987:
Hooray! Thanks!

PR #2765, opened 2025-04-02
Mlflow logging, checkpointing, and resuming

ChristophJud:
### MLFlow logging
- Added logging support to an MLFlow server
- For development purposes, also our custom JWT auth is supported for connecting to the server
- Basic model metrics are tracked

### Checkpointing
- Checkpoints are pushed to an S3 bucket
- A reference file is generated with a link to the S3 object which is then logged in MLFlow

### Resuming
- With the argument `mlflow_run_id` a particular run of an experiment can be resumed

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -47,7 +47,7 @@ checkpointing_bucket:
   type: str
   description: "S3 bucket name where the checkpoints are being written to."
 mlflow_run_id:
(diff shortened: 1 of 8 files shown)


ChristophJud:
Locally tested

ChristophJud:
my mistake sorry

PR #2834, opened 2025-06-05
CLI Catalog

vmiller987:
Hello!

I made this CLI Catalog when I first started using nnUNet. I've found it to be a handy reference material. I generated this by examining the entry point files. 

I believe this would be beneficial to any new users as it would give a simple location to examine all of the CLI arguments. For example, I shared the prediction section with another user (#2817 ) to answer his question.

If there's anything I need to fix, correct, or adjust, please let me know.

### DIFF
--- a/documentation/cli_catalog.md
+++ b/documentation/cli_catalog.md
@@ -1,0 +1,994 @@
+# CLI Command Catalog
+There are 21 commands for nnUNetv2.
+
+For basic use, there's only three commands that are needed:
+- `nnUNetv2_plan_and_preprocess`
+- `nnUNetv2_train`
+- `nnUNetv2_predict`
+
+# Available Commands
+- [CLI Command Catalog](#cli-command-catalog)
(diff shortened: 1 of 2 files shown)


vmiller987:
This is a bit out of date currently (or I missed a few arguments initially). If requested, I will update the catalog. 

I reference this quite often.

FabianIsensee:
My only issue with this is that this will inevitably get out of date when we do changes and forget to update :-/ Similar to how you already noticed yourself this is outdated ;-)
This is why I am hesitant to merge it...

vmiller987:
Would it be better to have it updated but have a "Last Updated: DD/MM/YYYY" block at the top? This will at least provide a reference material for other users. 

I made this manually awhile ago. Claude would excel at this simpler task. I will send my Claude subscription to fix/update this if desired. 

I understand the eventual out of date concern again. If you don't wish to include this, we can close this PR.

PR #2857, opened 2025-07-07
fix copying dataset.json to `output_folder` in `ensemble_folders` function

hxviet:
This fix changes how the dataset.json file is written to the output folder.

The old implementation tries to copy the file at the path `join(list_of_input_folders[0], 'dataset.json'`. When `dataset_json_file_or_dict` is specified, the file at this path is probably not intended to be used by `ensemble_folders` or may not even exist. The new implementation dumps the `dataset_json` object instead to ensure `ensemble_folders` writes the correct dataset.json data to the output folder whether `dataset_json_file_or_dict` is None, a file path, or a dictionary.

### DIFF
--- a/nnunetv2/ensembling/ensemble.py
+++ b/nnunetv2/ensembling/ensemble.py
@@ -3,6 +3,8 @@ import multiprocessing
 import shutil
 from copy import deepcopy
 from typing import List, Union, Tuple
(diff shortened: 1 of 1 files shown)


FabianIsensee:
Thanks!

PR #2861, opened 2025-07-14
Initialize predictor from individual path

MattTheCuber:
This PR adds the method `initialize_from_paths` as an alternative to `initialize_from_trained_model_folder` to the `nnUNetPredictor` class. With this method, instead of passing the entire training directory, you can just pass the files.

### DIFF
--- a/nnunetv2/inference/predict_from_raw_data.py
+++ b/nnunetv2/inference/predict_from_raw_data.py
@@ -128,6 +128,56 @@ class nnUNetPredictor(object):
             print('Using torch.compile')
             self.network = torch.compile(self.network)
 
+    def initialize_from_paths(self, dataset_path: str, plans_path: str, weights_path: str):
+        """
+        A more flexible method for initialize_from_trained_model_folder
+        """
+        dataset_json = load_json(dataset_path)
+        plans = load_json(plans_path)
+        plans_manager = PlansManager(plans)
+
+        checkpoint = torch.load(weights_path, map_location=torch.device('cpu'), weights_only=False)
+        trainer_name = checkpoint['trainer_name']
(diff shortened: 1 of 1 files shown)


FabianIsensee:
@claude please do a brief review of this. Seems useful!

FabianIsensee:
This PR seems to have some issues that would need to be resolved for merging. I would also like to understand why this is needed. It adds complexity and if in the future we decide to have different (or more) file we need to initialization we would need to change the signature of this function. nnU-Net puts everything in one folder already, so why not keep using that?

MattTheCuber:
My main reasoning was to reduce the number of files and file size of deploying models. We deploy our models to customers. Zipping up the results folder is pretty large. For example, one of our training output folders is 5.8 GB, where pulling out the three necessary files ends up being under 0.5 GB. This PR just makes extracting the files easier. A better solution might be to add a single ZIP/TAR file mode and some utilities for pulling relevant files from the training folder to create them. Then these could be deployed as a single file and read by nnUNet.

FabianIsensee:
You don't need to zip the entire output folder. You can create a custom script for zipping checkpoints that only includes the necessary files. Those should be the three files your PR links and nothing else (but preserve the folder structure)

FabianIsensee:
A small note on deploying to customers: You should delete the optimizer states from the checkpoints. This is a free halving of the checkpoint size :-)

MattTheCuber:
Our use case would find it easier not to have to run a script on a folder that creates a zip with a specific hierarchy, but I understand if this PR isn't the optimal solution. Mostly just wanted to give more flexibility in model loading rather than a specific directory structure with three files.

Your note would be very useful, what key is the optimizer states stored under?

FabianIsensee:
image

image

:-)

PR #2862, opened 2025-07-15
Fixing a bug in predict_from_files_sequential for 3d_cascaded_fullres models

helrewaidy:
This is to fix a bug in the predict_from_files_sequential() when predicting 3d cascaded models, where stacking the input image with the labels from a previous stage is missing. To resolve this issue, the labels from the prior stage were converted to one-hot and stacked with the input images and fed to the cascaded model.


FabianIsensee:
Good catch, thank you

Issue #2396, opened 2024-07-26
New error on windows: RuntimeError: Encountered inf in predicted array


wasserth:
Since yesterday I get the following error when running nnU-net v2.4.2 on windows. On ubuntu and mac this error does not happen. On windows this error also did not happen 3 days ago. Probably there was an update to some dependency without fixed version and this caused this new behaviour. Did you also come across this problem? Could it be a solution to clip inf values to the max number supported by the used dtype?

```
File "C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\site-packages\totalsegmentator\nnunet.py", line 251, in nnUNetv2_predict
    predictor.predict_from_files(dir_in, dir_out,
  File "C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\site-packages\nnunetv2\inference\predict_from_raw_data.py", line 256, in predict_from_files
    return self.predict_from_data_iterator(data_iterator, save_probabilities, num_processes_segmentation_export)
  File "C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\site-packages\nnunetv2\inference\predict_from_raw_data.py", line 373, in predict_from_data_iterator
    prediction = self.predict_logits_from_preprocessed_data(data).cpu()
  File "C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\site-packages\nnunetv2\inference\predict_from_raw_data.py", line 490, in predict_logits_from_preprocessed_data
    prediction = self.predict_sliding_window_return_logits(data).to('cpu')
  File "C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\site-packages\torch\utils\_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
  File "C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\site-packages\nnunetv2\inference\predict_from_raw_data.py", line 651, in predict_sliding_window_return_logits
    predicted_logits = self._internal_predict_sliding_window_return_logits(data, slicers,
  File "C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\site-packages\nnunetv2\inference\predict_from_raw_data.py", line 607, in _internal_predict_sliding_window_return_logits
    raise e
  File "C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\site-packages\nnunetv2\inference\predict_from_raw_data.py", line 600, in _internal_predict_sliding_window_return_logits
[...]

PR #2886, opened 2025-08-08
Fix Windows CI

dyollb:
Partially fixes issue #2396 

fix windows ci
- use torch==2.4.1 (see https://github.com/MIC-DKFZ/nnUNet/issues/2396#issuecomment-3162932241_
- activate windows action

normalize gaussian using numpy
- as noted [here](https://github.com/MIC-DKFZ/nnUNet/issues/2396#issuecomment-3163312603), normalization using torch produces 'Inf' for the test case with torch==2.4.0.
- doing the operation on the numpy array instead should not have any negative side-effects

specify folder 'nnunetv2' in pyproject.toml
- doesn't hurt. I noticed when I had temporary folders (e.g., 'venv/'), setuptools was not able to install the package properly
- add `.venv/` to the gitignore file. This is a common name for a virtual environment

note: this PR does not 'fix' the issue with 2.4.0. I tried accumulating the `predicted_logits` in predict_from_raw_data.py using float32, which also helped. But in the end I still got a dice of 0.008

### DIFF
--- a/.github/workflows/run_tests_nnunet.yml
+++ b/.github/workflows/run_tests_nnunet.yml
@@ -8,8 +8,7 @@ jobs:
   run-tests:
     strategy:
       matrix:
-        # os: [ubuntu-latest, windows-latest, macos-latest]  # fails on windows until https://github.com/MIC-DKFZ/nnUNet/issues/2396 is resolved
-        os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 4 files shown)


FabianIsensee:
Thanks! I merged this. Let's :crossed_fingers: that everything works now :-)

Issue #2863, opened 2025-07-15
Workers no longer alive when training on SLURM cluster


fervegl:
I am working on using nnUNET v2 for prostate cancer segmentation, the system I am using is a high-performance computing system with SLURM, however, whenever I leave the job working it often gets killed with the RunTimeError: "One or more background workers are no longer alive. Exiting. Please check the print statements above for the actual error"

This is the resources I am asking for job 
```
#SBATCH  --job-name=nnUNET
#SBATCH --gres=gpu:a100:1      
#SBATCH --cpus-per-task=1
#SBATCH --nodes=1
#SBATCH --mem=22GB                
#SBATCH --time=5-00:00:00         

export NCCL_P2P_DISABLE=1 
export nnUNet_n_proc_DA=1
export nnUNet_shm_used=0
export nnUNet_compile=False

nnUNetv2_train 16 3d_fullres 0 -device 'cuda' -p nnUNetResEncUNetXLPlans
```

Then once I get it running, the error shows up after a while, this is the log:
```
#######################################################################
Please cite the following paper when using nnU-Net:
Isensee, F., Jaeger, P. F., Kohl, S. A., Petersen, J., & Maier-Hein, K. H. (2021). nnU-Net: a self-configuring method for deep learning-based biomedical image segmentation. Nature methods, 18(2), 203-211.
#######################################################################

2025-07-15 01:37:08.465818: do_dummy_2d_data_aug: True
2025-07-15 01:37:08.492321: Using splits from existing split file: /tmp/nnUNet_preprocessed/Dataset015_PICAI/splits_final.json
2025-07-15 01:37:08.497755: The split file contains 5 splits.
2025-07-15 01:37:08.499371: Desired fold for training: 1
2025-07-15 01:37:08.500742: This split has 1181 training and 295 validation cases.
Exception in thread Thread-1 (results_loop):
Traceback (most recent call last):
  File "/cvmfs/hpc.rug.nl/versions/2023.01/rocky8/x86_64/amd/zen3/software/Python/3.11.5-GCCcore-13.2.0/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "/cvmfs/hpc.rug.nl/versions/2023.01/rocky8/x86_64/amd/zen3/software/Python/3.11.5-GCCcore-13.2.0/lib/python3.11/threading.py", line 975, in run
    self._target(*self._args, **self._kwargs)
[...]

Issue #2825, opened 2025-05-28
Background workers not alive / EOF error / no CUDA GPU available


stang29:
Hi there, 

I saw a few threads previously about some of these errors but none have seemed to work, was wondering if I could get some assistance with these errors! I've attached my requirements.txt file, the specs of the server with gpu I have to use (four 32Gb Tesla V100 = 128Gb total)  and pasted the script I use to run nnunet. My images are 256^3 and I'm working in a python virtual environment. I've tried different combinations of setting those environment variables listed below but depending on which I use, I get one of either Background workers not alive error, no CUDA GPU available error (even though I check before I run the script that it can see / use cuda), EOF error or a strange keyboard interrrupt error. I'm not using a docker container, although if that is recommended please let me know. 

I've been stuck on this for a while, any suggestions is appreciated!

[cudainfo.txt](https://github.com/user-attachments/files/20478588/cudainfo.txt)

[requirements.txt](https://github.com/user-attachments/files/20478505/requirements.txt)

Script:

# export OMP_NUM_THREADS=1
# export CUDA_VISIBLE_DEVICES=1
# export TORCHDYNAMO_DISABLE=1
#export OPENBLAS_NUM_THREADS=1
#ulimit -u 1540468
#export TORCHDYNAMO_VERBOSE=1

export nnUNet_raw="/m/Researchers/usr/deeppvs/for_nnunet/nnUNet_raw"
export nnUNet_preprocessed="/m/Researchers/usr/deeppvs/for_nnunet/nnUNet_preprocessed"
export nnUNet_results="/m/Researchers/usr/deeppvs/for_nnunet/nnUNet_results"

nnUNetv2_plan_and_preprocess -d 001 -pl nnUNetPlannerResEncL --verify_dataset_integrity

export nnUNet_n_proc_DA=24

nnUNetv2_train 001 3d_fullres 5 -p nnUNetResEncUNetLPlans --npz

FabianIsensee:
Hey all, are you still experiencing this issue? What is the full traceback? The `Background workers not alive` is usually accompanied by some additional error that points to the actual problem. Usual suspects are insufficient shared memory (/dev/shm), most often encountered in docker containers, or insufficient system RAM

Issue #2777, opened 2025-04-08
About nondet_multi_threaded_augmenter


a-die:
Hello, I rewrote a multi-task learning framework for nnUNet. In short, in addition to the segmentation task, there is also a task similar to learning the connectivity between image pixels. I rewrote the dataset, dataloader, and trainer, mainly adding another additional label reading. These are the background.
The problem is that when I use multi-threaded dataloader, it seems to cause memory leaks, that is, using the code:
![Image](https://github.com/user-attachments/assets/78713653-55a3-4d50-815c-ecad933423ea)
mt_gen_train = LimitedLenWrapper(self.num_iterations_per_epoch, data_loader=dl_tr, transform=tr_transforms,
num_processes=allowed_num_processes, num_cached=6, seeds=None,
pin_memory=self.device.type == 'cuda', wait_time=0.02)
And the error is as follows:
Traceback (most recent call last):
File "/home/i/miniconda3/envs/nnUNet/bin/nnUNetv2_train", line 8, in <module>
sys.exit(run_training_entry())
File "/home/i/ASOCA/nnUNet/nnunetv2/run/run_training.py", line 290, in run_training_entry 
run_training(args.dataset_name_or_id, args.configuration, args.fold, args.tr, args.p, args.pretrained_weights,
 File "/home/i/ASOCA/nnUNet/nnunetv2/run/run_training.py", line 211, in run_training
 nnunet_trainer.run_training()
 File "/home/i/ASOCA/nnUNet/nnunetv2/training/nnUNetTrainer/nnUNetTrainer.py", line 1276, in run_training
 train_outputs.append(self.train_step(next(self.dataloader_train)))
 File "/home/i/miniconda3/envs/nnUNet/lib/python3.10/site-packages/batchgenerators/dataloading/nondet_multi_threaded_augmenter.py", line 196, in __next__
 item = self.__get_next_item()
 File "/home/i/miniconda3/envs/nnUNet/lib/python3.10/site-packages/batchgenerators/dataloading/nondet_multi_threaded_augmenter.py", line 181, in __get_next_item
 raise RuntimeError("One or more background workers are no longer alive. Exiting. Please check the "
RuntimeError: One or more background workers are no longer alive. Exiting. Please check the print statements above for the actual error message
But there is no specific error message.
However, there is no problem using SingleThreadedAugmenter (dl_tr, tr_transforms), but the running speed is very slow.
Can anyone give me some ideas for troubleshooting?

[...]

Issue #2769, opened 2025-04-02
Training RuntimeError: One or more background workers are no longer alive


nrepina:
Hello, I am trying to run training with the command:

nnUNetv2_train 3 2d 0 --npz

in python 3.10, on the Nvidia Tesla A100 GPU, on our computing cluster requesting 48GB memory. I've tried decreasing the batch size of the dataset to 10. This should certainly be enough memory for a 2D training (I see max memory usage for the job is 2.2GB). However the job keeps erroring out within a minute of starting the training run. I see this is a common issue, and unfortunately this means I cannot use nnUnet for training. Any advice would be much appreciated. 

Please see logs below:

-------
2025-04-02 15:31:28.731068: The split file contains 5 splits.
2025-04-02 15:31:28.743055: Desired fold for training: 0
2025-04-02 15:31:28.752051: This split has 13 training and 4 validation cases.
using pin_memory on device 0
Max memory usage in bytes: 2230145024

[INFO] [2025-04-02T15:31:43+02:00] [1549827] Workflow finished with code 
[INFO] [2025-04-02T15:31:43+02:00] [1549827] Workflow execution time \(seconds\) : 71

----------
Error logs:
-----------
Exception in thread Thread-3 (results_loop):
Traceback (most recent call last):
  File "/path/Python/mamba-envs/nnunet/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
Traceback (most recent call last):
  File "/path/Python/mamba-envs/nnunet/bin/nnUNetv2_train", line 8, in <module>
    sys.exit(run_training_entry())
  File "/path/Python/mamba-envs/nnunet/lib/python3.10/site-packages/nnunetv2/run/run_training.py", line 267, in run_training_entry
    run_training(args.dataset_name_or_id, args.configuration, args.fold, args.tr, args.p, args.pretrained_weights,
  File "/path/Python/mamba-envs/nnunet/lib/python3.10/site-packages/nnunetv2/run/run_training.py", line 207, in run_training
    nnunet_trainer.run_training()
  File "/path/Python/mamba-envs/nnunet/lib/python3.10/site-packages/nnunetv2/training/nnUNetTrainer/nnUNetTrainer.py", line 1363, in run_training
    self.run()
  File "/path/Python/mamba-envs/nnunet/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
[...]

PR #2910, opened 2025-09-23
Fix for "One or more background workers are no longer alive. Exiting" errors

MikeUU332:
Hello,

We encountered the "One or more background workers are no longer alive. Exiting" error as described in:

https://github.com/MIC-DKFZ/nnUNet/issues/2863
https://github.com/MIC-DKFZ/nnUNet/issues/2825
https://github.com/MIC-DKFZ/nnUNet/issues/2777
https://github.com/MIC-DKFZ/nnUNet/issues/2769
https://github.com/MIC-DKFZ/nnUNet/issues/2766
https://github.com/MIC-DKFZ/nnUNet/issues/2749

when running nnUNet on an HPC cluster. A colleague and I looked into this issue and came across this link:
https://github.com/joblib/threadpoolctl/issues/176

which says that using `with threadpool_limits` is not thread safe and to use a global mutex instead. `with threadpool_limits` is used in both nnUNet and batchgenerators libraries. By using a global mutex instead of with threadpool limits as done in this pull request, we no longer encounter this issue. We verified that nnUNet output is similar (they can't be identical with the non-determinative multi-threaded launcher). 

We encountered this error with both nnUNet and the batchgenerators library so a similar pull request will be done for batchgenerators.

### DIFF
--- a/nnunetv2/training/dataloading/data_loader.py
+++ b/nnunetv2/training/dataloading/data_loader.py
@@ -217,6 +217,7 @@ class nnUNetDataLoader(DataLoader):
                         seg_all = torch.stack(segs)
                     del segs, images
                     global_mutex.release()
+                
             return {'data': data_all, 'target': seg_all, 'keys': selected_keys}
 
         return {'data': data_all, 'target': seg_all, 'keys': selected_keys}


FabianIsensee:
Apologies I initially thought this was a duplicate to https://github.com/MIC-DKFZ/batchgenerators/pull/136 but it's not. My bad.
Same issues as in batchgenerators though. This PR removes the threadpoolctl call which is something we should not do.
I suspect that this is a system misconfiguration rather than a bug. Probably insufficient shared memory (/dev/shm) or system RAM. I would like to be able to reproduce this if somehow possible

FabianIsensee:
Somewhere someone suggested to uncomment the line `# multiprocessing.set_start_method("spawn")` in run_training.py. Can you please try to see if that fixed your issue?

MikeUU332:
Hello, I tried different spawn methods before but they did not fix the problem. Further usage of nnunet showed this issue happening on one of our clusters and not on a different one, although they have identical configurations in all ways we can think of to check. Based on the threadpool link previously mentioned, a colleague suggested running nnUnet with Python's faulthandler enabled, which resulted in

Fatal Python error: Segmentation fault
Current thread 0x0000148dbfa30740 (most recent call first):
File "~/nnunet/lib/python3.12/site-packages/threadpoolctl.py", line 1017 in _find_libraries_with_dl_iterate_phdr
…
So the crash is in threadpoolctl which led us to follow the suggestion in [(https://github.com/joblib/threadpoolctl/issues/176)] and add the global mutex, which worked for us. As mentioned there, this definitely seems to be a hit or miss issue, which are always the worst to track down.

FabianIsensee:
Well it didn't resolve the issue within threadpoolctl, it just removed threadpoolctl from the equation :-/ Interesting finding though. Is the threadpoolctl version identical between the systems?

PR #2926, opened 2025-10-24
FIX: print logs when training early end due to an exception when training with nnUNetTrainerBenchmark_5epochs and nnUNetTrainerBenchmark_5epochs_noDataLoading

Leirbag-gabrieL:
Hi,

First time using nnUNet and I encountered unexpected behavior while trying to do a dry run of my code. 
The training early ended while using benmarking trainers and said `fastest_epoch = 'Not enough VRAM!'` :/

While actually, there were something else going on (I have a 40 GB GPU but 8GB config doesn't fit in ?)
The lack of error logs made me waste a bit of time because I did not know what was actually wrong.
The logging implementation was hiding this error:

```
backend='inductor' raised:
RuntimeError: Triton Error [CUDA]: device kernel image is invalid

Set TORCH_LOGS="+dynamo" and TORCHDYNAMO_VERBOSE=1 for more information


You can suppress this exception and fall back to eager by setting:
    import torch._dynamo
    torch._dynamo.config.suppress_errors = True
```
   
For now, I did not fix my issue, but now it is correctly reported!

### DIFF
--- a/nnunetv2/dataset_conversion/Dataset015_018_RibFrac_RibSeg.py
+++ b/nnunetv2/dataset_conversion/Dataset015_018_RibFrac_RibSeg.py
@@ -43,10 +43,10 @@ if __name__ == '__main__':
         seg_file = join(base, c + '-label.nii.gz')
         if not isfile(seg_file):
             # test case
-            shutil.copy(img_file, join(imagesTs, c + '_0000.nii.gz'))
+            shutil.copyfile(img_file, join(imagesTs, c + '_0000.nii.gz'))
             continue
         n_tr += 1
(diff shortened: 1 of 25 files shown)


FabianIsensee:
Hey Gabriel, thanks for opening this PR. Unfortunately it seems to contain a collection of changes from your personal branch that are not related to the issue you mentioned above. If there are individual changes you would like to see integrated into nnU-Net, please create dedicated PRs for each of them. Ideally discuss this beforehand with us in the issues so that we can give our input!
Best,
Fabian

Leirbag-gabrieL:
MB I will redo my PR, I thought I created a separate branch.

Issue #2927, opened 2025-10-24
Calculate `num_iterations_per_epoch` dynamically in `nnUNetTrainer`


mohamedbassem6:
In `nnUNetTrainer.py` [line 149](https://github.com/MIC-DKFZ/nnUNet/blob/86606c53ef9f556d6f024a304b52a48378453641/nnunetv2/training/nnUNetTrainer/nnUNetTrainer.py#L149), the number of iterations per epoch is hardcoded to be 250. I believe a more correct approach would be to calculate it dynamically based on the batch size, and the number of GPUs.

FabianIsensee:
Dear @mohamedbassem6 this is intentional and has been a design staple of nnU-Net since the beginning. We do not use epochs in the classical sense where one epoch is an iteration over the whole dataset. This is not even possible in patch-based segmentation unless you precompute patch locations and then iterate over all patches (again something we are not doing). Changing this would have unintended/unclear consequences for nnU-Net's segmentation performance

PR #2928, opened 2025-10-24
Fix: calculate num_iterations_per_epoch dynamically in nnUNetTrainer

mohamedbassem6:
**Fixes:** #2927

This update calculates the number of iterations per epoch dynamically instead of hardcoding it to 250.

The method `get_tr_and_val_datasets` in the `nnUNetTrainer` class was modified to update both `self.num_iterations_per_epoch` and `self.num_val_iterations_per_epoch` based on the dataset size and batch size.

Additionally, progress bars were added to the training and validation loops to provide users with a clearer indication of progress during long training sessions.


FabianIsensee:
As outlined in #2927 this is intended behavior and not a bug. We do not want to change it. Thank you for raising this discussion though!

PR #2939, opened 2025-11-12
Allow passing kwargs to image reader/writer

GFabien:
Hi,
Thanks for the great repository, it is super useful!
I started adding new image reader writers to try stuff with my data and I found it practical to allow passing kwargs to my new image reader writers instead of creating many classes. Here is a small PR to implement that. Tell me if it's okay for you. The idea is just to add an optional key to the dataset.json called "image_reader_writer_kwargs" and store the kwargs there.
Thank you,
Fabien


FabianIsensee:
Dear Fabien, thanks for this PR! Can you please explain when this is needed? A concrete example would be great!

Issue #2911, opened 2025-09-25
Adding a new `nnUNetTrainer` variant for extensive data augmentations on `GPU`


NathanMolinier:
## Description

Hello nnUNet Community,

I am currently working on a project exploring how data augmentations can improve model performance for 3D segmentation tasks on MRI images, using nnUNet as a baseline for my experiments.

As part of this effort, I introduced a new trainer called `nnUNetTrainerDAExt` (shared in another [issue](https://github.com/MIC-DKFZ/nnUNet/issues/2733#issue-2910733621), and potentially a future PR), which incorporates a broader range of augmentations and has shown promising improvements during inference.

However, the main drawback of this trainer—as with other augmentation-based trainers in the [variants](https://github.com/MIC-DKFZ/nnUNet/tree/master/nnunetv2/training/nnUNetTrainer/variants) folder—is that augmentations are performed on the CPU. This creates a significant bottleneck, especially when combining multiple augmentations, while the GPUs remain underutilized. For instance, with `nnUNetTrainerDAExt`, depending on the number of CPU cores available (since transforms use multiprocessing), one epoch can take anywhere between 700 seconds and over 3000 seconds—leading to training times that can stretch beyond a month.

To address this issue, I developed a [new GPU-based augmentation trainer](https://github.com/neuropoly/AugLab/blob/8dd65c49186fd79458372256a53a4bfabcc306bb/auglab/trainers/nnUNetTrainerTest.py#L135). In my tests, this trainer achieved a 3× reduction in epoch time compared to the CPU-based version (from ~700s down to ~260s on my cluster).

At the moment, this trainer doesn’t yet include as many augmentation options as [nnUNetTrainerDAExt](https://github.com/MIC-DKFZ/nnUNet/issues/2733#issue-2910733621), but I am actively working on expanding it—either by implementing new transforms or by integrating existing ones from libraries such as [Kornia](https://github.com/kornia/kornia) or [torchvision](https://docs.pytorch.org/vision/0.9/transforms.html). Admittedly, finding 3D transforms is more challenging than for 2D.

If anyone is interested in contributing to the development of this trainer, your help would be greatly appreciated!

## Related issues

- https://github.com/MIC-DKFZ/nnUNet/issues/1453

etienne87:
[...]

PR #2944, opened 2025-11-18
Feat/gpu augmentations

etienne87:
In an effort for GPU Augmentation (See https://github.com/MIC-DKFZ/nnUNet/issues/2911#issuecomment-3533240137), i add an option "gpu_augmentation", and i added a new iterator called "ThreadedGPUAugmenter" that applies data augmentation after transfer to the GPU.


FabianIsensee:
Dear Etienne,
thanks for this PR! The current data augmentation pipeline is not GPU compatible. I think this would require major additional effort with limited benefits as one usually has sufficient CPU power to run data augmentation in the background, and even if not then there still is the possibility of using SegOrd0 variants of the trainers to cut down on CPU utilization. Can you please outline situations in which this approach offers substantial advantages over the current workflow, given that integrating this would require changes to batchgenerators to make everything GPU compatible and overall increase complexity?
Thanks a lot!
Best,
Fabian

etienne87:
Hello Fabian! 

This MR is indeed a big change, probably not worth in the nominal cases. I see improvements when CPU is busy with other trainings. People get crowded on one server, and everybody uses the CPU for augmentation, and each GPU is starving for data.

In our team everybody uses nnUNetv2 which already switched to nearest neighbor if i am not mistaken ? 

At the same time, this GPU-only solution introduces its own caveats as you mentioned in the other [MR](https://github.com/MIC-DKFZ/batchgeneratorsv2/pull/15):

- Some time of the GPU is now dedicated to data augmentation
- Requires the introduction of this [new class ](https://github.com/MIC-DKFZ/nnUNet/pull/2944/changes#diff-ab21f147767b5e90b6720e94d7b9782763c3b853fb48d94bc37c29c2572e8472) because multiprocessing seem to fail when patches are too big to be shared via the Queue (since i rotate on the gpu raw patches need to be bigger to avoid padding values) 

Realistically, I agree that other, simpler solutions can fix our problems:

- accelerating grid_sample on the CPU (pytorch seems to use only 1 thread for all voxels for grid_sample c++ cpu side) 
- use cheaper interpolation as you say.

Anyway, 
it has been some time since i worked on this topic, but if it can help, i can come back with raw numbers on some specific training configurations.

FabianIsensee:
Hey etienne,
it's similar on our cluster. nnU-Net-style workloads (nnU-Net + its variants) make up a huge fraction of jobs. But we don't encounter CPU issues and if we are, we simply switch to nearest neighbor interpolation for segmentation augmentation (nnUNetTrainer_DASegOrd0) to address that. I  would very much prefer to keep things as they are, since switching to GPU augmentation would really have only marginal benefits for a small group of users while increasing the complexity by a lot
Best,
Fabian

etienne87:
ok, agreed, closing for now.

PR #2992, opened 2026-02-27
fix: remove unused imagecodecs dependency

### LINKED ISSUE #2899: What is the purpose of the `imagecodecs` package in `nnunetv2`'s requirements?
Hi there! We are a downstream consumer of `nnunetv2`. We recently ran into issues with the `imagecodecs` package due to the maintainer deleting wheel files from PyPI, which broke frozen requirements:

- https://github.com/cgohlke/imagecodecs/issues/129

I was curious why `imagecodecs` was getting installed in the first place, and found that `nnunetv2` has installed this package since the commit https://github.com/MIC-DKFZ/nnUNet/commit/fee6fa2baec15a08bb35f1c579edfbe4a0e72383. But, there are no other [code results, no imports, no usages](https://github.com/search?q=repo%3AMIC-DKFZ%2FnnUNet+imagecodecs&type=code). 

So, what purpose does `imagecodecs` have for the `nnunetv2` package? Is it necessary?

VAIDEHI-28:
Fixes #2899

The `imagecodecs` package was listed as a dependency in `pyproject.toml` 
but is not imported or used anywhere in the nnunetv2 codebase.

This caused installation failures for users when wheel files became 
unavailable on PyPI.

This PR removes the unused dependency to prevent future breakage.

Verified by:
- Searched entire codebase for any `imagecodecs` references → zero results
- Successfully installed and ran nnUNet without `imagecodecs`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,6 @@ dependencies = [
     "nibabel",
     "matplotlib",
     "seaborn",
-    "imagecodecs",
     "yacs",
     "batchgeneratorsv2>=0.3.0",
     "einops",


FabianIsensee:
`imagecodecs` is used by tifffile to en- and decode compressed tiff files. This is a required dependency which (for reasons I don't understand) is not immediately installed with tiffile. Can you please explain when exactly you encounter installation issues?

PR #3005, opened 2026-03-20
fix: add actionable fix instructions to affine/origin/direction mismatch warnings

### LINKED ISSUE #2751: Encountered an error when preprocessing LiTS dataset with nnunetv2 2.5.2
Hello @FabianIsensee and the team,

I've just encountered an error below when preprocessing LiTS dataset with nnunetv2 2.5.2

**Command:**
```
nnUNetv2_plan_and_preprocess -d 001 -c 3d_fullres --verify_dataset_integrity
```

**Error log:**
```
WARNING: Affine is not the same for image and seg! 
Affine image: [[  -0.83593798   -0.           -0.          208.8999939 ]
 [  -0.            0.83593798   -0.         -213.1643219 ]
 [   0.           -0.            2.5        -632.75      ]
 [   0.            0.            0.            1.        ]] 
Affine seg: [[1. 0. 0. 1.]
 [0. 1. 0. 1.]
 [0. 0. 1. 1.]
 [0. 0. 0. 1.]]
Image files: ['./raw/Dataset001_LiTS/imagesTr/LiTS_053_0000.nii']. 
Seg file: ./raw/Dataset001_LiTS/labelsTr/LiTS_053.nii.
This can be a problem but doesn't have to be. Please run nnUNetv2_plot_overlay_pngs to verify if everything is OK!

Traceback (most recent call last):
  File "/home/mtoan65/anaconda3/envs/nnunetv252/bin/nnUNetv2_plan_and_preprocess", line 8, in <module>
    sys.exit(plan_and_preprocess_entry())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mtoan65/anaconda3/envs/nnunetv252/lib/python3.11/site-packages/nnunetv2/experiment_planning/plan_and_preprocess_entrypoints.py", line 180, in plan_and_preprocess_entry
    extract_fingerprints(args.d, args.fpe, args.npfp, args.verify_dataset_integrity, args.clean, args.verbose)
  File "/home/mtoan65/anaconda3/envs/nnunetv252/lib/python3.11/site-packages/nnunetv2/experiment_planning/plan_and_preprocess_api.py", line 47, in extract_fingerprints
    extract_fingerprint_dataset(d, fingerprint_extractor_class, num_processes, check_dataset_integrity, clean,
  File "/home/mtoan65/anaconda3/envs/nnunetv252/lib/python3.11/site-packages/nnunetv2/experiment_planning/plan_and_preprocess_api.py", line 30, in extract_fingerprint_dataset
    verify_dataset_integrity(join(nnUNet_raw, dataset_name), num_processes)
  File "/home/mtoan65/anaconda3/envs/nnunetv252/lib/python3.11/site-packages/nnunetv2/experiment_planning/verify_dataset_integrity.py", line 219, in verify_dataset_integrity
    raise RuntimeError(
RuntimeError: Some images have errors. Please check text output above to see which one(s) and what's going on.
```

VAIDEHI-28:
Fixes #2751

## Problem
When running `nnUNetv2_plan_and_preprocess --verify_dataset_integrity`, 
the origin, direction, and affine mismatch warning messages were vague 
and gave users no context about what the warning meant or how serious it was.

## Fix
Improved the warning messages in `verify_dataset_integrity.py` to:
- Clearly state that these are warnings only and will not cause a crash
- Remind users to verify alignment visually using `nnUNetv2_plot_overlay_pngs`
- Avoid making any assumptions about how to fix the geometry mismatch
  (as the correct fix depends entirely on how the data was prepared)

## Before
Warning: Origin mismatch between segmentation and corresponding images.
Origin images: (-100.0, 150.0, 50.0).
Origin seg: (0.0, 0.0, 0.0).
No context, no guidance ❌

## After
Warning: Origin mismatch between segmentation and corresponding images.
Origin images: (-100.0, 150.0, 50.0).
Origin seg: (0.0, 0.0, 0.0).
This is a warning only and will not cause a crash. Please verify that
your segmentation and image are correctly aligned by running:
nnUNetv2_plot_overlay_pngs
Clear, informative, no unsafe assumptions ✅

## Verification
Tested with a dummy CT dataset with intentionally mismatched affines 
warnings now clearly communicate severity and next steps without 
making any assumptions about the user's data.

### DIFF
--- a/nnunetv2/experiment_planning/verify_dataset_integrity.py
+++ b/nnunetv2/experiment_planning/verify_dataset_integrity.py
@@ -88,14 +88,12 @@ def check_cases(image_files: List[str], label_file: str, expected_num_channels: 
         affine_image = properties_image['nibabel_stuff']['original_affine']
         affine_seg = properties_seg['nibabel_stuff']['original_affine']
         if not np.allclose(affine_image, affine_seg):
-            print('WARNING: Affine is not the same for image and seg! \nAffine image: %s \nAffine seg: %s\n'
+            print('Warning: Affine mismatch between segmentation and corresponding images. '
+                '\nAffine image: %s \nAffine seg: %s\n'
                 'Image files: %s. \nSeg file: %s.\n'
-                'How to fix: Copy the image header to the segmentation using nibabel:\n'
-                '    import nibabel as nib\n'
-                '    img = nib.load("your_image.nii.gz")\n'
-                '    seg = nib.load("your_segmentation.nii.gz")\n'
-                '    seg_fixed = nib.Nifti1Image(seg.get_fdata(), img.affine, img.header)\n'
-                '    nib.save(seg_fixed, "your_segmentation_fixed.nii.gz")\n'
+                'This is a warning only and will not cause a crash. Please verify that your '
+                'segmentation and image are correctly aligned by running: '
+                'nnUNetv2_plot_overlay_pngs\n'
                 % (affine_image, affine_seg, image_files, label_file))
 
     # sitk checks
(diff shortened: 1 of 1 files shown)


VAIDEHI-28:
"Note: The Codespell check failure is caused by pre-existing typos in other files unrelated to this PR (Dataset015_018_RibFrac_RibSeg.py, export_prediction.py, nnunet_logger.py, etc.). My changes pass codespell cleanly. Happy to fix those in a separate PR if needed!"

VAIDEHI-28:
That's a really fair point, thanks for raising that!

Thinking about this more carefully, I think a simple small/large split might still be too blunt. Here's a more nuanced approach I had in mind:

1. **Negligible difference (< 1mm)** - almost certainly a floating point rounding error from file format conversion (e.g. DICOM → NIfTI pipelines are known to introduce sub-millimeter origin drift). Safe to suggest CopyInformation with a clear disclaimer that the user should only do this if they are confident the segmentation was created on this image.

2. **Moderate difference (1mm – 10mm)** - ambiguous. Could be rounding or could be a real shift. Don't suggest any automated fix — just direct the user to visually verify in ITK-SNAP or with nnUNetv2_plot_overlay_pngs before doing anything.

3. **Large difference (> 10mm)** - almost certainly a real geometric mismatch. Issue a strong warning that this is likely a data preparation error and the segmentation may have been created on a different image entirely. No fix suggested just tell the user to go back and check their data pipeline.

On top of this, I was also thinking about something a bit more proactive - instead of only looking at the header numbers, we could do a quick sanity check on the actual voxel data itself. Specifically, check whether the segmentation foreground voxels (anything > 0) fall within the non-zero region of the image. If the headers are mismatched but the foreground of the segmentation still overlaps with the image content, it's a strong signal that the data is probably fine and the mismatch is just a metadata issue. If the foreground falls completely outside the image's non-zero region, that's a red flag that something is genuinely wrong with the data and we should warn much more aggressively regardless of how small the header difference looks.

This way we're not just comparing numbers we're actually checking whether the data is clinically usable, which feels like a more robust signal.

Does this direction make sense ?
Eagerly waiting for your guidance on this.

PR #3007, opened 2026-03-26
Extended Metrics Beyond Dice and IoU in Summary Reporting [HD95, RVE, ASSD]

toufiqmusah:
- Extend evaluation metrics to include ASSD, HD95, and signed RVE alongside Dice/IoU.
- Compute spacing-aware surface distances and respect ignore labels.
- Print new metrics in final validation summary output.
- Metrics flow into [summary.json] and foreground/class means automatically.
- Final validation now prints Dice, ASSD, HD95, and RVE.


FabianIsensee:
@claude let's try this again

FabianIsensee:
Awesome, that went better than expected. I like all the suggestions by claude. Particularly the correctness of the HD95 implementation should be checked and I would really want these advanced metrics to be opt in due to their high computational overhead.
For RVE we should discuss what the intended behavior for missing classes should be (np.nan vs np.inf) and what poential effects this may have for aggregation. What are your thoughts on this @toufiqmusah ?
If we report RVE, why not also include AVE (absolute volume error) as well? Might be useful for some cases and is easy enough to add.
In the same spirit: NSD@1mm might also be a sensible addition given that the Metrics Reloaded project recommends that in conjunction with Dice? The NSD implementation by Deepmind also uses surface distances, so there is some potential for synergies

Best,
Fabian

toufiqmusah:
Hello @FabianIsensee,

thanks for taking the time to look at this. Interesting use of claude too, given the amount of thankless maintenance you do on the repo. Isn't it expensive though? I agree with the findings overall. This would be better as an opt-in due to computational overheads, and defaulting to np.inf may cause some aggregation issues. How about `np.nanmean` instead of `np.nan`?

I will work on addressing the challenges over the weekend, and explore the suggested metrics (including NSD, AVE). My current implementation heavily borrows from the deepmind surface-distance implementations, by the way. 

Best,
Toufiq

FabianIsensee:
Hey Toufiq,
we are just trying claude to see how much it helps us. If it saves a lot of time, the cost is worth it. It takes about 0.25 EUR per response in the current setup. If you compare that to what one of us maintainers costs, it can be a good deal. But like I said, we are trying it out and will evaluate once we have some data :-)
If np.inf are present in RVD then the aggregation is unclear. np.inf is a valid output (FP or FN prediction of a class) and it should be up to the user to decide how they handle it. They still get per-case metrics in the summary.json so they can figure this out themselves. np.nanmean would be the correct aggregation scheme here. The user would in the presence of np.inf get np.inf as aggregated result and that should trigger them thinking about how they want to hande it.
Honestly, I personally don't like volume metrics. They can only ever be used as supplementary information and never standalone because they do not check whether the predicted volume is actually what we wanted to segment, so using something like the Dice is mandatory.
BTW the same aggregation issue also arises for FP/FN predictions with ASSD and HD95. It is unclear how they should behave in these cases. Aggregation becomes meaningless because you can't determine a distance to something that does not exist. HD95/ASSD only  really make sense for structures that you would expect to be present and predicted in each image, like organs. I would always pick NSD over these for pathologies becauyse NSD handles missing pred/gt gracefully.
Best,
Fabian

aymuos15:
Just wanted to chime in here and say coderabbit (what monai uses) is free for opensource. 

Also, just to chime in again, lol, I would personally not want surface metrics by default. Would defintely want it through a part of custom trainer per say. Mainly because they are not gpu friendly.

PR #3008, opened 2026-03-29
Improve environment variable validation across core workflows

Pasupuleti-12:
## Summary
This PR improves how nnU-Net validates required environment variables for path-based workflows.

## Problem
Missing path environment variables can currently surface later as less actionable failures. Raising immediately at import time would also break workflows that do not require all three paths.

## What this PR changes
- adds reusable helpers in `nnunetv2.paths` for validating required environment variables
- preserves import-time behavior so optional workflows are not broken by missing unrelated paths
- adds contextual validation where the paths are actually required:
  - experiment planning / preprocessing
  - training
  - model import / export / download
  - output folder resolution

## Why this approach
This keeps the change generally applicable across setups while avoiding unnecessary breakage in workflows that only depend on a subset of nnU-Net paths.

## Validation
- verified updated modules compile successfully with `python -m compileall`
- manually verified warning behavior on import when env vars are missing
- manually verified explicit errors when required paths are missing
- manually verified success behavior when paths are set

### DIFF
--- a/nnunetv2/experiment_planning/plan_and_preprocess_api.py
+++ b/nnunetv2/experiment_planning/plan_and_preprocess_api.py
@@ -8,7 +8,10 @@ from nnunetv2.configuration import default_num_processes
 from nnunetv2.experiment_planning.dataset_fingerprint.fingerprint_extractor import DatasetFingerprintExtractor
 from nnunetv2.experiment_planning.experiment_planners.default_experiment_planner import ExperimentPlanner
 from nnunetv2.experiment_planning.verify_dataset_integrity import verify_dataset_integrity
-from nnunetv2.paths import nnUNet_raw, nnUNet_preprocessed
+from nnunetv2.paths import (
+    require_preprocessed_dataset_path,
+    require_raw_dataset_path,
+)
 from nnunetv2.utilities.dataset_name_id_conversion import convert_id_to_dataset_name
 from nnunetv2.utilities.find_class_by_name import recursive_find_python_class
 from nnunetv2.utilities.plans_handling.plans_handler import PlansManager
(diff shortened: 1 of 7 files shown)


FabianIsensee:
@claude your thoughts on this?

FabianIsensee:
Thanks for the PR @Pasupuleti-12 ! I will need to think about this. I am wondering if there is a simpler way of ensuring the environment variables are set when they are needed

Pasupuleti-12:
Thanks for taking a look, @FabianIsensee.

I agree with simplifying this. My goal was to ensure missing env vars fail at actual usage points, while avoiding import-time hard failures for workflows that do not require all paths.

Given your note, I can narrow this PR to a smaller approach:
- keep import-time behavior unchanged (no warning semantics change),
- add explicit checks only at clear workflow entry points,
- avoid partial broad helper rollout across modules,
- fix fragile transitive import usage and remove redundant checks.

This should keep behavior consistent and reduce maintenance overhead. If this direction matches what you prefer, I’ll push a reduced patch accordingly.

FabianIsensee:
Codex, with some nudging from my end, found a very elegant way of doing this via a custom _EnvPath class. Take a look at master :-)

Pasupuleti-12:
Thanks for the review and guidance, @FabianIsensee.

I can see the `custom_EnvPath` approach now on `master`, and it is a cleaner solution than my current patch. I’m closing this PR as superseded and will align future contributions with this direction.

I appreciate the feedback.

FabianIsensee:
Apologies for not accepting the PR as it :see_no_evil: I know you spend a lot of time on that and appreciate the effort!

Pasupuleti-12:
Thank you for the feedback and for taking the time to review my PR  I really appreciate it.

I understand the decision and would love to improve based on your suggestions. If possible, could you please share specific areas where my approach or implementation can be improved?

I’m very interested in contributing further to this project and have also submitted a GSoC proposal aligned with this work. Any guidance or feedback on how I can better align my contributions with the project expectations would be extremely helpful.

Looking forward to contributing again!

FabianIsensee:
Nothing inherently wrong with your approach. It required calling a dedicated function each time the variables were used. That was a bit clunky. The new approach doesn't circumvent that entirely (at least not when running ==) but at least most usages are handled automatically. os.path.join for example triggers the __str__ (or __fspath__) function which call require()

Pasupuleti-12:
Thanks a lot for the detailed explanation  that really helps me understand the design preference better.

I see how leveraging Python’s built-in behaviors like `__str__`/`__fspath__` makes the validation more seamless and avoids the need for explicit calls everywhere. That’s a much cleaner approach.

I’d love to explore improving my implementation in that direction. Would you recommend extending the current abstraction to better integrate with these automatic conversions, or are there specific areas in the codebase where such improvements would be most impactful?

Looking forward to contributing further!

FabianIsensee:
I am a big proponent of not changing a running system unless there is a clear case for why it's needed. Right now there are no major issues or pending reworks that need to be taken care of. Thanks for you offer though!

PR #3020, opened 2026-04-14
fix: logging implementation was hidding exceptions in nnUNetTrainerBenchmark_5epochs and nnUNetTrainerBenchmark_5epochs_noDataLoading

Leirbag-gabrieL:
Hi,

The logging implementation in both edited file was hiding this error:

```
backend='inductor' raised:
RuntimeError: Triton Error [CUDA]: device kernel image is invalid

Set TORCH_LOGS="+dynamo" and TORCHDYNAMO_VERBOSE=1 for more information
You can suppress this exception and fall back to eager by setting:
    import torch._dynamo
    torch._dynamo.config.suppress_errors = True
```

It was reporting this error as :
fastest_epoch = 'Not enough VRAM!'

While I had more than enought VRAM to run my network.

I got the solution for my error here : https://github.com/pytorch/pytorch/issues/119054

Apparently something is wrong with the ptxas bundled by default with triton. 
Using my own, by defining the environment variable TRITON_PTXAS_PATH="/usr/local/cuda-11.6/bin/ptxas"  fixed my issue.

I changed the logging so that it displays the exception in the logs and now the message is accurate :
fastest_epoch = 'Training has ended due to an error. Check your logs for more information.'

### DIFF
--- a/nnunetv2/training/nnUNetTrainer/variants/benchmarking/nnUNetTrainerBenchmark_5epochs.py
+++ b/nnunetv2/training/nnUNetTrainer/variants/benchmarking/nnUNetTrainerBenchmark_5epochs.py
@@ -27,9 +27,10 @@ class nnUNetTrainerBenchmark_5epochs(nnUNetTrainer):
     def run_training(self):
         try:
             super().run_training()
-        except RuntimeError:
+        except RuntimeError as e:
             self.crashed_with_runtime_error = True
             self.on_train_end()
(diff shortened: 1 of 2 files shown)


FabianIsensee:
Thanks for this isolated PR! We should not call on_train_end when the training crashed. Is there a specific reason this is there?

FabianIsensee:
I see that I myself added this in one of the trainers. I need to check why, brb

FabianIsensee:
Thanks, looks good!

PR #3027, opened 2026-04-27
Fixed typo in saving properties associated to predicted probabilities

### LINKED ISSUE #2781: apply_inference_nonlin after averaging probabilities
Hi, 
I am trying to understand how nnUNet ensembling is performed and I'm wondering about the use of `label_manager.convert_logits_to_segmentation` which internally calls `self.apply_inference_nonlin` after averaging the probabilities.
Here's what I came up with:

- When we call `nnUNetv2_predict` with `--save_probabilities` inference non-linearity layer is applied and what is stored in the `.npz` files are probabilities and not logits
- Probabilities coming from different models are averaged
- `label_manager.convert_logits_to_segmentation` is called instead of `label_manager.convert_probabilities_to_segmentation` because after averaging in the multiclass case, you can have the sum of the probabilities for a voxel that is not 1. To ensure this, you apply once again the inference non-linearity layer. 

Can you confirm that this is the right thinking?
In a binary case, reapplying the sigmoid layer doesn't change the results ?

TimZaragori:
Fixes #2781 

Typo in the `merge_files` function when saving the averages probabilities with the `--save_probabilities` option in `nnUNetv2_predict` for example.

`probabilities` were save in the pickle file instead of `properties`

### DIFF
--- a/nnunetv2/ensembling/ensemble.py
+++ b/nnunetv2/ensembling/ensemble.py
@@ -44,7 +44,7 @@ def merge_files(list_of_files,
     image_reader_writer.write_seg(segmentation, output_filename_truncated + output_file_ending, properties)
     if save_probabilities:
         np.savez_compressed(output_filename_truncated + '.npz', probabilities=probabilities)
-        save_pickle(probabilities, output_filename_truncated + '.pkl')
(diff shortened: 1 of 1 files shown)


FabianIsensee:
Thank you!

Issue #3010, opened 2026-04-02
Advice on training strategy: Large dataset (>6000 images),


valdemarbang:
Hello!

We are currently working on our Master's Thesis and are dealing with a strict time limit and a large dataset of over 6,000 annotated images.

We are experimenting with 2 different models (nnLandmark and nnMamba) across 6 different dataset structures (adding various segmentations as additional channels). Because of the time limit of the thesis for training all these variations, we are unable to use the full 6,000 images or standard 5-fold cross-validation for every experiment. 

**Current Setup & Observations**

In our first iteration to save time, we trained on a smaller subset:
- Split: 720 Train / 180 Val / 100 Test
- Setup: Fold 0 only, default 1000 epochs.

We noticed two main issues during this run:

1. Early Convergence: The Validation New best EMA pseudo Dice is consistently found around epoch 400-500. 
2. Unrepresentative Test Set: The purely random 100-image test set lacked enough "hard" cases. As a result, the model's accuracy on the test set was artificially high and not a reliable indicator of real-world performance.

For the second iteration we added 1000 completely new unseen test images to get more hard cases to our dataset.
- Split: 800 Train / 200 Val / 1000 Test
- Setup: Fold 0 only, 500 epochs. 

**Proposed Solutions & Questions**

Given our limited compute time, we are considering a few paths forward and would love your advice on best practices within the nnU-Net framework:

1. Is it advisable to shorten the training to 500 epochs?
Since the pseudo Dice peaks early, we are considering to set self.num_epochs = 500 and continuing to train only on Fold 0. This would allow us to rapidly test all 12 model/dataset combinations.

2. What is the optimal workflow for our final evaluation?
Due to time constraints, we cannot train a 5-fold ensemble for all dataset structures. Would the best approach be to use Fold 0 (with 500 epochs) as a rapid study to find the most promising model and dataset structure, and then run a full 5-fold cross-validation solely on that winning combination? Or should we skip the 5-fold completely (skip validation dataset) and only use Training and Test dataset?

Thank you in advance for your time and any guidance you can provide!

Issue #2754, opened 2025-03-26
--val_best during train based on val_loss or psuedo-dice?


rbramkumar:
I am new to this implementation so I'm wondering whether the "--val_best" option does what I think it does. In my use-case, it looks like the val loss stops decreasing quite early. Do you know if the checkpoint model saves the best model according to "val_loss" (red line) or "pseudo-dice" (green line). Is there a way to save the best model according to the evaluation metric i.e., green line? 

Also, putting it out there in case anyone knows more than me here - any suggestions to bring down val_loss? 

`!nnUNetv2_train 555 2d 0  -p nnUNetResEncUNetMPlans --val_best`

![Image](https://github.com/user-attachments/assets/4ae30d9e-7e6f-4c63-ac69-902edc76e58f)

constantinulrich:
Hi, we calculate an exponential moving average on the pseudo dice for the "best" checkpoint. But take that with care. We do not do a proper validation - its just a pseudo dice with 50 foreground patches.

constantinulrich:
> Also, putting it out there in case anyone knows more than me here - any suggestions to bring down val_loss?

There are trainers implemented that use a stronger data augmentation scheme (nnUNetTrainerDA5), which you could test...

rbramkumar:
Thank you very much. I will try out the trainer with data augmentations. 

Thanks,
Ram

PR #3030, opened 2026-04-30
Improve --val_best

Mts027:
Hey nnU-Net Team,

I have been using nnU-Net extensively, thank you for providing and maintaining such an awesome project!

# The Problem

nnU-Net tends to overfit on some datasets. To help with this, one can sometimes use `--val_best`, but as I understand it, this is not generally recommended ([1](https://github.com/MIC-DKFZ/nnUNet/issues/2754#issuecomment-2766015379), [2](https://github.com/MIC-DKFZ/nnUNet/blob/2026b8e0652886508a75df0b2f5b95dd252bec54/nnunetv2/run/run_training.py#L233), [3](https://github.com/MIC-DKFZ/nnUNet/issues/3010#issuecomment-4206273449)).

This is because:

1. It uses pseudo Dice, which is a biased metric, since it samples based on `oversample_foreground_percent`.
2. Since this signal is noisy, it needs smoothing, which makes the value "lag". However, nnU-Net's training, and thus model performance, is inherently noisy from epoch to epoch.

# The Idea

The idea here is to replace nnU-Net's default validation strategy by selecting a fixed subset of all sliding-window tiles from the final validation, and then evaluating on that fixed subset. This tries to address:

1. the sampling bias by using fair random sampling, and
2. the noisy signal by making the validation subset comparable from epoch to epoch.

# Experiments

I ran these experiments on datasets where I had already trained models for 1000 epochs in ([4](https://arxiv.org/pdf/2511.17146)). These use the `noSmooth` trainer variants, but that should not matter too much. The table below shows the Dice scores for both runs: the old default run and the new run using the proposed `val_best` strategy.

| Dataset | Old run | Old run | New run | New run | Old Δ | New Δ |
|---|---:|---:|---:|---:|---:|---:|
|  | Final | Best | Final | Best | Best - Final | Best - Final |
| WMH | 77.11 | 78.06 | 78.46 | 80.15 | +0.95 | **+1.69** |
| SBM | 67.49 | 68.31 | 66.40 | 69.25 | +0.82 | **+2.85** |

These runs are not perfectly comparable, since nnU-Net training is not deterministic. Still, the new best-model selection shows larger within-run gains than the previous strategy: +1.69 and +2.85 Dice points compared to +0.95 and +0.82.

I am of course always open to discussion, so let me know what you think! This will need more extensive evaluation on different datasets (any experiments welcome! :)) and different model architectures. For example, I think the PrimusV2/V3 trainers could also greatly benefit from this, since I have noticed them having a stronger tendency to overfit.

Kind regards,

Luc

Disclaimer: Parts of the code in this PR were written with the help of Codex.

### DIFF
--- a/nnunetv2/training/nnUNetTrainer/fixed_val_tiles.py
+++ b/nnunetv2/training/nnUNetTrainer/fixed_val_tiles.py
@@ -1,0 +1,162 @@
+from dataclasses import dataclass
+from itertools import product
+from typing import List, Optional, Tuple
+
+import numpy as np
+import torch
+from acvl_utils.cropping_and_padding.bounding_boxes import crop_and_pad_nd
+
+from nnunetv2.inference.sliding_window_prediction import compute_steps_for_sliding_window
+
+
+@dataclass(frozen=True)
+class FixedValTile:
+    """A deterministic validation crop.
+
+    key is the dataset case identifier passed to dataset.load_case.
+    starts can have these shapes:
+    - (y_start, x_start) for native 2D patches on 2D cases.
+    - (z_start, y_start, x_start) for native 3D patches on 3D cases.
+    - (slice_idx, y_start, x_start) for 2D patches applied slice-by-slice to 3D cases.
+    """
+    key: str
+    starts: Tuple[int, ...]
+
+
+@dataclass(frozen=True)
+class FixedValTileStats:
+    num_cases: int
+    num_candidate_tiles: int
+    num_requested_tiles: int
+    num_selected_tiles: int
+    num_tiles_on_rank: int
+
+
+class FixedValTileManager:
+    def __init__(self, dataset, patch_size: Tuple[int, ...], batch_size: int, transforms,
+                 tile_step_size: float, seed: int, num_tiles: Optional[int], default_num_tiles: int,
+                 is_ddp: bool = False, rank: int = 0, world_size: int = 1):
+        self.dataset = dataset
+        self.patch_size = patch_size
+        self.batch_size = batch_size
+        self.transforms = transforms
+        self.tile_step_size = tile_step_size
+        self.seed = seed
+        self.num_tiles = num_tiles
+        self.default_num_tiles = default_num_tiles
+        self.is_ddp = is_ddp
+        self.rank = rank
+        self.world_size = world_size
+        self.tiles, self.stats = self._build_tiles()
+
+    def __len__(self):
+        return len(self.tiles)
+
+    def iter_batches(self):
+        for batch_start in range(0, len(self.tiles), self.batch_size):
+            yield self._make_batch(self.tiles[batch_start:batch_start + self.batch_size])
+
+    def _build_tiles(self) -> Tuple[List[FixedValTile], FixedValTileStats]:
+        tiles = []
+        num_cases = len(self.dataset.identifiers)
+
+        for key in self.dataset.identifiers:
+            data, _, _, _ = self.dataset.load_case(key)
+            image_size = tuple(int(i) for i in data.shape[1:])
+
+            if len(self.patch_size) < len(image_size):
+                # 2D configurations keep 3D cases as (c, z, y, x) but use 2D patches (y, x), so we
+                # enumerate in-plane sliding-window tiles for each slice independently.
+                assert len(self.patch_size) == len(image_size) - 1
+                # compute_steps_for_sliding_window requires image_size >= patch_size. For smaller validation
+                # slices we enumerate one virtual patch-sized tile and let crop_and_pad_nd add the padding later.
+                tiled_size = tuple(max(i, p) for i, p in zip(image_size[1:], self.patch_size))
+                steps = compute_steps_for_sliding_window(tiled_size, self.patch_size, self.tile_step_size)
+                for slice_idx in range(image_size[0]):
+                    for starts in product(*steps):
+                        tiles.append(FixedValTile(key, (slice_idx, *starts)))
+            else:
+                # compute_steps_for_sliding_window requires image_size >= patch_size. For smaller validation
+                # images we enumerate one virtual patch-sized tile and let crop_and_pad_nd add the padding later.
+                tiled_size = tuple(max(i, p) for i, p in zip(image_size, self.patch_size))
+                steps = compute_steps_for_sliding_window(tiled_size, self.patch_size, self.tile_step_size)
+                for starts in product(*steps):
+                    tiles.append(FixedValTile(key, tuple(starts)))
+
+        num_candidate_tiles = len(tiles)
+        requested_num_tiles = self.default_num_tiles if self.num_tiles is None else int(self.num_tiles)
+        selected_num_tiles = min(requested_num_tiles, len(tiles))
+
+        if selected_num_tiles < len(tiles):
+            rng = np.random.default_rng(self.seed)
+            selected_indices = np.sort(rng.choice(len(tiles), size=selected_num_tiles, replace=False))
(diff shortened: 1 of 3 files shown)


FabianIsensee:
Thanks, I will take a look next week. My main concern is that it just overfits more to the given validation set. This is what your experiment shows. Do you have any evidence that this type of checkpoint selection benefits an independent test set?

Mts027:
Hey, that's a good point!

So I thought about this and, for me, there are 2 use cases and 4 methods to use nnU-Net:

Use cases:

1. You care about **absolute** performance, e.g., you are using nnU-Net on a new type of problem.
2. You care about **relative** performance, e.g., you are comparing X network architectures or loss functions against each other.

Methods: 

1. You do 5-fold CV with 1000 epochs and report that score.
2. You do 5-fold CV with 1000 epochs but with `--val-best`.
3. You do 5-fold CV with 1000 epochs and report on a test set.
4. You do 5-fold CV with 1000 epochs with `--val-best` and report on a test set.

For use case 1:

Method 2 is obviously not correct: that is what my experiments were showing, as you could be influencing your validation scores. So if you don't have an external test set, I don't think you should be using `--val-best`, but with an external test set, you could consider using `--val-best`, as long as I/we can show that it improves test performance.

For use case 2:

This is a little bit more complicated: when choosing between methods 1 and 2, this is very use case dependent. For example, if you are comparing multiple setups inside the nnU-Net framework, you could argue `--val-best` gives each setup the "fairest" training schedule, because some methods might overfit later than others or not at all. Now you could also argue the opposite and say that you are favoring setups with high variance in performance, since you might just be selecting models that have random high spikes in validation performance, which is also a fair point. And, for me at least, when selecting between the two, there is no clear winner, and you can argue for both sides.

So when doing these experiments, I think the question we want to answer is: Does `--val-best` improve test performance and, if so, does my PR improve it more than the current version? This would address case 1, and for case 2, it would suggest that if test performance improves as well, the gains in validation performance aren't just overfitting.

I used the same WMH dataset from above, but replaced SBM with the ACDC dataset, since I don't have the test labels for SBM. I ran everything for 1000 epochs and 5-fold CV using [this trainer](https://github.com/Mts027PRs/nnUNet/blob/6d0bd08bd7a731849a82c39bf6a878cca904d762/nnunetv2/training/nnUNetTrainer/variants/sampling/nnUNetTrainerFixedValTiles.py), which allows me to save `final`, `best`, and `best_old` checkpoints. 

<table><thead><tr><th rowspan="2">Dataset</th><th colspan="3">Validation Dice</th><th colspan="3" style="border-left:2px solid">Test Dice</th></tr><tr><th>Final</th><th>Best old</th><th>Best</th><th style="border-left:2px solid">Final</th><th>Best old</th><th>Best</th></tr></thead><tbody><tr><td>WMH</td><td>0.7729</td><td>0.7852</td><td><strong>0.7946</strong></td><td style="border-left:2px solid">0.7775</td><td>0.7872</td><td><strong>0.7917</strong></td></tr><tr><td>ACDC</td><td>0.9191</td><td>0.9206</td><td><strong>0.9212</strong></td><td style="border-left:2px solid">0.9200</td><td>0.9208</td><td><strong>0.9222</strong></td></tr></tbody></table>

All conclusions here will be premature, because we are only using two datasets, and I would love some external tests! (probably on datasets with higher inter-method variance)

I think the most interesting conclusion is: both `--val-best` setup tend to increase performance on these test sets, but this PR does it slightly better. I also don't see any signs of overfitting on the validation set. Now, the gains are very small, so the conclusion here is very far from definitive. 

Let me know what you think!

FabianIsensee:
OK convinced that this is something we can include. Experiments are sound and the changes are relatively small.
But I need to take a closer look at the code. As codex always does, it introduced a bunch of extra stuff and classes that make life harder rather than easier. And it does not understand how partial reading works (indexing with [:] reads and decompresses the entire image...). I think this might have to be reworked for simplicity. Will get back to you on this next week

FabianIsensee:
One issue we need to think about is that we are currently saving checkpoints by trainer, config and plans name. We might need to make sure it is traceable what val_best setting was being used and name checkpoints accordingly

Mts027:
Hey, thanks for taking a look :)

- I think I fixed the `[:]` access pattern.
- To be fair to Codex, that extra class was my design :sweat_smile:. I didn't want to add even more LOC to `nnUNetTrainer.py`. I don't really have a strong opinion on this, so feel free to make design changes as you wish!
- As for the checkpoint naming: I don't know how comfortable you would be with this, but I would suggest replacing the old val-best behavior instead of making this an option, since it's really the same idea, just without the noise added by the old design, which I don't think serves any real purpose. This would also solve the naming issue, since we'd only have one val-best option left.
- There are also a few other design decisions where I need input, such as the fixed seed.

FabianIsensee:
I would not want to overwrite the old val-best behavior because here we are not really looking at the entire validation set and thus effectively use a smaller val set than the old behavior

Mts027:
First, just to be clear: I’m definitely less familiar with nnU-Net internals than you 😅 So please correct me if I’m misunderstanding something here.

My current understanding from the experiments is that, with the very limited statistical significance of only two runs, the proposed `val_best` behavior seems to select checkpoints with better final validation Dice than the current method. But I agree that the tested datasets have relatively few validation (sliding) windows, so the fixed subset covers most, or  basically all, of the validation set. For large datasets this is not true anymore. For example, KiTS has over 20k validation windows, so 100 fixed windows is clearly only a small subset.

So yes, this is a weak point of the proposed method.

However, I think the current behavior has a different weak point. It may look at many / all validation windows over training, but not for the same checkpoint. Since it samples 100 windows per epoch and smooths pseudo Dice with an EMA, it assumes that model performance is roughly stable across multiple epochs.

I think this can be dangerous, depending on the dataset. In the fixed-tiles SBM run, the same 100 windows are evaluated every epoch, so we can measure how far the EMA-smoothed pseudo Dice is from the pseudo Dice at epoch `t`. This is probably an optimistic estimate, because in the current behavior the sampled windows also change from epoch to epoch.

For fold 0, over the last 100 epochs:

```text
mean |pseudo - EMA| = 0.007554
p95  |pseudo - EMA| = 0.023558
max  |pseudo - EMA| = 0.029401
```

So even late in training, the EMA can be a few Dice points away from the current checkpoint’s pseudo Dice in the tail cases.

This matters because the current `val_best` selects the maximum EMA value over training. If the EMA is still noisy, taking the maximum tends to select an extreme/lucky value. So while the current method may cover more of the validation set over time, the selected checkpoint can still be the one with a particularly lucky EMA trajectory.

Over the last 500 epochs this gets even stronger:

```text
p95 |pseudo - EMA| = 0.053409
max |pseudo - EMA| = 0.099457
```

And since checkpoint selection is done over all epochs, not only the final 100, this “selecting the lucky maximum” issue can become even worse.

I also don’t think just changing the EMA decay fully fixes this. Lower decay reacts more to recent observations, but then fewer windows effectively contribute to the estimate. Higher decay smooths more, but then it depends on a longer history of previous checkpoints, which makes the temporal-stability assumption even stronger.

So I don’t think either behavior is perfect. The old behavior has better coverage over time, but only by relying on this temporal-stability assumption. The proposed fixed-subset behavior avoids that temporal mixing and makes checkpoint-to-checkpoint comparisons cleaner, but it can clearly suffer from subset bias on large datasets.

Now the new behavior still has issues, as you said. It only selects a subset, and if you are unlucky that subset might be a bad representation of the problem space. I’m honestly not sure what the right solution should be here, but maybe there is some middle ground, like using fixed tiles with more windows and validating only every few epochs to avoid the extra overhead.

Let me know what you think!

PR #3033, opened 2026-05-09
[FIX] patch sampling in get_bbox() method to ensure patches are always fully contained inside the volume when oversampling foreground

sifaoso:
Hey all,

Thanks a lot for all the work that has gone into nnU-Net over the years!

I believe I found a bug in patch sampling when visualizing 3D CT batches (AMOS dataset), where some sampled patches contain slices outside the scan field-of-view (resulting in unexpected padding).

The issue occurs in the `get_bbox()` method when `force_fg=True` (foreground oversampling). In this case, a foreground class is first selected, then a random voxel from that class is chosen as the patch center. However, when constructing the patch, the resulting bounding box is only constrained with respect to the lower bounds (`lbs`) and is not checked against the upper bounds (`ubs`), which define the valid sampling range to keep the patch fully inside the volume (here: https://github.com/MIC-DKFZ/nnUNet/blob/e73bbc743f93877211ca434dd07eca07c1221d0b/nnunetv2/training/dataloading/data_loader.py#L145). As a consequence, some foreground patches may partially fall outside the valid image region and become padded.

Note that this issue does not occur when `force_fg=False`, since `bbox_lbs` is sampled directly within the valid interval `[lbs, ubs]`.

The proposed fix explicitly constrains the sampled bounding box to remain within both `lbs` and `ubs`, which fixes the issue in my tests.

Below is a small GIF illustrating the problem (see the 3rd sample: some slices lie outside the scan field-of-view and are therefore padded unexpectedly).

batches-nnunet-pr

### DIFF
--- a/nnunetv2/training/dataloading/data_loader.py
+++ b/nnunetv2/training/dataloading/data_loader.py
@@ -142,7 +142,7 @@ class nnUNetDataLoader(DataLoader):
                 # selected voxel is center voxel. Subtract half the patch size to get lower bbox voxel.
                 # Make sure it is within the bounds of lb and ub
                 # i + 1 because we have first dimension 0!
-                bbox_lbs = [max(lbs[i], selected_voxel[i + 1] - self.patch_size[i] // 2) for i in range(dim)]
+                bbox_lbs = [max(lbs[i], min(ubs[i], selected_voxel[i + 1] - self.patch_size[i] // 2)) for i in range(dim)]
             else:
                 # If the image does not contain any foreground classes, we fall back to random cropping
                 bbox_lbs = [np.random.randint(lbs[i], ubs[i] + 1) for i in range(dim)]


FabianIsensee:
Thanks for the report. Looks legit! I am in vacation right now and will look into it once I am back

sifaoso:
Hey @FabianIsensee - any update on this by any chance?

FabianIsensee:
Experiments testing whether this has an impact are now running. Will get back with results on Thursday

FabianIsensee:
Hey, results are in. I also ran a suite of experiments with 'unbounded' meaning the constraint on the selected voxel was lifted entirely

image

It seems like none of this makes a difference. The results are all within regular run-to-run variance. This is fold_0 only.
I need to think about this some more. We either need to bound lower and upper coordinates, or not bound coordinates at all. Not entirely sure which is the better approach tbh. You have any thoughts?

sifaoso:
Thanks for the experiments! I was expecting these kinds of results. It is true that it's difficult to draw a conclusion from now... Personally, I find it more intuitive to ensure that the patches are fully contained within the scans (i.e., no padding) by bounding lower and upper coordinates. I believe this also depends on the inference strategy, and, in particular, on how the sliding window inference (SWI) is implemented. I need to look at it a bit more in detail.
But my point is that if SWI does not require any padding (e.g., the last window is shifted so that it still fits entirely within the volume), then the network will never encounter padded regions at test time. In that case, it makes sense to constrain the patches so that they lie entirely within the scan. 
The crucial point is that we should not reject boundary foreground voxels, but this is not the case. Therefore, we should probably favour a patch sampling strategy that is consistent with the SWI implementation. I'm curious about your thoughts on that?

FabianIsensee:
True but this is where things become tricky. What if in a dataset images sometimes are smaller than the patch size and need to be padded for inference? In that case the network sees empty borders and therefore it might be beneficial had it seen that during training as well. Normal sized images which are larger than the patch size will not have this issue and SWI will only show image patches

FabianIsensee:
I see that the current way of selecting the voxel is not ideal because we only bound one side. We either need to pick the voxel completely freely or completely bounded

sifaoso:
Yes, exactly. My PR is intended to enforce this behavior, ensuring that the center voxel is fully bounded, i.e., the extracted 3D patch is always entirely contained within the 3D volume.

I understand your point regarding cases where the image is smaller than the patch size and therefore needs to be padded during inference. In such situations, the padded regions can be removed afterward to obtain a clean segmentation restricted to the actual ROI.

However, in AMOS (and likely in other datasets as well), there are variations in FOV across scans, particularly along the z-axis. For example, some scans extend only to the bottom of the lungs, whereas others include the entire lungs. If the center voxel is allowed to be sampled freely, the model may receive patches that contain valid bottom lung slices and then suddenly padded (black) slices, while for others scans it may receive patches containing the entire lungs. Maybe, this padding could introduce discrepancies during training.

Below are some visualizations of 3D patches from AMOS comparing the bounded case (where the center voxel is fully bounded) and the unbounded case (where the center voxel is sampled freely within the foreground) for the same patches. For visualization purposes, I used `oversample_foreground_percent = 1.0`. Batch size is 2.

You have much more experience than me to take the final decision. Ensuring that patches are always fully contained within the scan might probably be the safest option, but I completely agree that we should carefully consider potential edge cases.

**[BOUNDED SAMPLING] (proposed PR)**

https://github.com/user-attachments/assets/562cd1cf-905d-45f3-98ef-53d60861899a


https://github.com/user-attachments/assets/2a3f494a-a59a-4b1d-8525-fb5a1523dbbe

**[UNBOUNDED SAMPLING]**


https://github.com/user-attachments/assets/c6a0f4c6-84f8-4c02-bd45-539706c87398


https://github.com/user-attachments/assets/dca1c870-16d9-44ee-8612-2cf725cc9bdd

FabianIsensee:
I know what you mean. My main issue which makes it hard to decide. If an AMOS test case has less than 192 slices and you train wiht a patch size of 193^3 then the image needs to be padded, to at inference time the model will see padded data. So the question is whether it would not be beneficial for the model to also see that during trainig

PR #3039, opened 2026-06-05
Add official PrimusV2 and PrimusV3 trainers

TaWald:
Add PrimusV2-S/B/M/L and PrimusV3-S (also B/M/L, which so far have not been too useful).


FabianIsensee:
Thanks can you please document which of these (V2/V3) corresponds to the new Primus paper?

PR #3047, opened 2026-07-02
Fix UnboundLocalError when passing a dict to load_checkpoint

vcasellesb:
### Description
This PR fixes a bug in `nnUNetTrainer.load_checkpoint` where passing a checkpoint as a dictionary (as permitted by the `Union[dict, str]` type hint) results in an `UnboundLocalError`. 

Previously, if a dictionary was passed, the `isinstance(filename_or_checkpoint, str)` block was skipped, leaving the local variable `checkpoint` undefined before the subsequent `for` loop attempted to iterate over `checkpoint['network_weights']`.

### The Fix
Renamed the input argument from `filename_or_checkpoint` to `checkpoint`. 
* If a `str` is passed, it loads the file and correctly overwrites the `checkpoint` variable.
* If a `dict` is passed, it bypasses the string check and seamlessly falls through to the rest of the function using the provided dictionary.

This fix was applied to both the base `nnUNetTrainer.py` and the overridden method in `nnUNetTrainer_warmup.py`.

Note:
This description was written by Gemini and me.

### DIFF
--- a/nnunetv2/training/nnUNetTrainer/nnUNetTrainer.py
+++ b/nnunetv2/training/nnUNetTrainer/nnUNetTrainer.py
@@ -1217,12 +1217,12 @@ class nnUNetTrainer(object):
             else:
                 self.print_to_log_file('No checkpoint written, checkpointing is disabled')
 
-    def load_checkpoint(self, filename_or_checkpoint: Union[dict, str]) -> None:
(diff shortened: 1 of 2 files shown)


vcasellesb:
> 🔍 **nnU-Net Code Review**

Have also changed your minor suggestion.

FabianIsensee:
Thanks!

PR #3049, opened 2026-07-03
Harden trainer lookup against phantom module import failures

SID-6921:
## Summary
- guard built-in trainer discovery in 
ecursive_find_trainer_class_by_name against ModuleNotFoundError caused by phantom trainer modules reported by pkgutil
- if that specific import failure happens in 
nunetv2.training.nnUNetTrainer.*, continue to external trainer lookup via 
nUNet_extTrainer instead of aborting
- add regression test that simulates this failure and verifies successful fallback to an external trainer class

## Why
In some Python 3.12+ environments, module discovery can report modules that are not importable under the expected dotted path. This can abort built-in trainer probing before external trainer fallback is attempted.

## Scope
- Real dependency import failures are still propagated (only the phantom built-in module case is tolerated).

## Tests
- python -m pytest nnunetv2/tests/test_find_objects.py -q (5 passed)

### DIFF
--- a/nnunetv2/tests/test_find_objects.py
+++ b/nnunetv2/tests/test_find_objects.py
@@ -80,6 +80,42 @@ class TestFindObjects(unittest.TestCase):
 
             self.assertIn("definitely_missing_dependency", str(exc.exception))
 
+    def test_external_trainer_lookup_after_builtin_phantom_module_error(self):
+        with TemporaryDirectory() as trainer_dir:
+            _write_file(
+                os.path.join(trainer_dir, "ext_trainer.py"),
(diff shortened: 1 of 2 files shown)


FabianIsensee:
Can you take a look at the key findings of claude and address them?

SID-6921:
Addressed the key findings in 7130e1a. I removed the broad outer ModuleNotFoundError catch in recursive_find_trainer_class_by_name so genuine built-in import failures are no longer masked. The phantom-module handling remains in the inner recursive module loader where it belongs, and the related tests were updated accordingly.

Issue #3069, opened 2026-09-16
[Feature Request]: tiff compression of saved predictions


oceanites:
I use nnUnet for prediction of 2D-EM images. They are very large with edge lengths of between 30-40.000 pixels. Therefore, the resulting prediction tif files are also huge (ca. 2GB), but very sparse. Using a simple compression like PACKBITS compresses these files to ca. 20-40MB.

1. Is there already a flag to have nnunet write compressed tif outputs?
2. If not, would you welcome a PR implementing this? 
Having a short look, I think that the 2D tif writes are done using scikit image, which doesn't allow setting tiff settings seperately
https://github.com/MIC-DKFZ/nnUNet/blob/ded2aa3a4c81a9caae37054224d8ebac2d19a061/nnunetv2/imageio/natural_image_reader_writer.py#L65
Importing tifffile in that file (which is already used in the project) and having a switch based on "output_fname ends with tif/tiff" would be the most simple implementation. But also putting all of the 2D tif handling into a seperate ReaderWrite would be possible.

PR #3071, opened 2026-09-22
Compress 2D TIFF predictions with PackBits

yellowgram:
## Summary

Addresses https://github.com/MIC-DKFZ/nnUNet/issues/3069

Prediction pixels are unchanged. `NaturalImage2DIO.write_seg` still casts with `seg[0].astype(np.uint8 if np.max(seg) < 255 else np.uint16, copy=False)` and then writes that array. Only the TIFF container changes: `.tif` and `.tiff` (any case, including `pathlib.Path`) are stored with PackBits via `tifffile.imwrite`. PNG and BMP still use `skimage.io.imsave`.

Existing `Tiff3DIO` plans stay on zlib. This PR does not change `Tiff3DIO.write_seg`.

`.tiff` disambiguation matches `.tif`. `NaturalImage2DIO` is registered first, so a 2D `.tiff` example selects that reader. A 3D `.tiff` whose last axis is not 3 or 4 fails the 2D reader and falls through to `Tiff3DIO` when an example file is provided. Last-axis length 3 or 4 is still treated as RGB by the 2D reader; that ambiguity already exists for `.tif`.

PackBits stays the 2D default.

## Test

`nnunetv2/tests/test_natural_image_tiff_compression.py` — 9 tests, including PackBits tag/size, Path-safe extensions, uint16 `read_seg` round-trip, BMP header, and registry selection for 2D vs 3D `.tiff`.

Addresses https://github.com/MIC-DKFZ/nnUNet/issues/3069


yellowgram:
Thanks for the review.

Addressed in b4f9539:

- Added a changelog note under *Changes that affect custom trainers* covering (1) PackBits for all `.tif`/`.tiff` 2D writes (lossless; byte stream changes for downstream tools) and (2) `determine_reader_writer_from_file_ending('.tiff', None)` now returning `NaturalImage2DIO` — pass an example file for 3D disambiguation.
- Clarified the `write_seg` comment: PackBits (RLE) fits sparse row-major 2D label maps and is faster than Deflate; 3D stays zlib in `Tiff3DIO`.
- Softened the size assert to `compressed_size < uncompressed_size` (kept `compressed_size < expected.nbytes`).

Left the no-example-file test gap as documentation-only given all in-tree call sites supply an example file.

FabianIsensee:
Thank you!

